PR #9, opened 2022-07-11
Add files via upload

angehung5:
input_variables.txt updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /canopy
+output*.txt
 
 # Prerequisites
 *.d


zmoon:
This is a pretty big file to have in the repo, especially if it is going to keep being updated.

zmoon:
It looks like the test run is failing because `nlat` has changed, now 251 (25--50 N)

angehung5:
Sorry I forgot to change settings. The original input_variables.txt with 25-65N was too big and github didn't let me upload, so I changed domain to 25-50N. And I agree that it is a big file. I will change it to a smaller domain and add one more variable to the list.

zmoon:
Thanks @angehung5 , sounds good. If ~ 1 MB or less is possible that would be good, also speeding up the run time of the test.

drnimbusrain:
@angehung5 @zmoon 

Yes, I agree the test 2D input_variables.txt file should be smaller subset for say a highly vegetated area like the northeast or southeast for testing, and one could/should always test with a single line file representing a single lat/lon point to maintain column capability.   

Once the input_variables.txt file is updated for this and Z0 I will make a new canopy_parm module that takes the input grid variables and sets the canopy shape profile/parameters etc for calculations.   I will then add simple parameterization modules for canopy photolysis and turbulence attenuation.

zmoon:
Note: continued in #10

PR #16, opened 2022-07-25
Develop

drnimbusrain:


zmoon (commented):
This looks fine to me.

To ease maintainability, I think the description could be detailed in just one place (e.g. in the namelist file like you have, and/or in the namelist-reading module) and then more simplified elsewhere, like just "PAI calculation option"

canopy_parm_mod.F90
```diff
@@ -74,7 +74,8 @@ SUBROUTINE CANOPY_PARM( VTYPE, FCH, FFRAC, LAI, &
                 else if (PAI_OPT .eq. 2) then !need PAI function of model LAI
                     PAI=LAI
                 else
-                    write(*,*)  'No PAI_OPT Selected...Defaulting to option 0'
+                    write(*,*)  'Wrong PAI_OPT choice in namelist...exiting'
+                    call exit(123)
```
zmoon, line 78: Intrinsic `exit` subroutine is not part of the Fortran standard.

Also, just curious, why 123?

drnimbusrain: @zmoon Agreed. Just simply exiting with exit status omitted.  We can use stop instead.

zmoon: `stop` with no code could make the program exit with status code 0, which would be not desired. But `stop 123` should be fine.

zmoon: I could also see leaving it using the `exit` subroutine, since it is unlikely we would use other compilers.

drnimbusrain: Well yes, lets do that.  I think exit should work with other compilers as well.  I will switch to call exit(2) if you agree.

zmoon: Sounds ok, just thought I would note it, as I was reading [this thread](https://stackoverflow.com/a/34476302) about it the other day.

zmoon (approved):
Would still suggest simplifying the description of this new option in some places to something shorter and more general like "PAI calculation option"

PR #17, opened 2022-08-11
Develop

drnimbusrain:


zmoon (approved):
Looks fine but I'm not in love with the name `pai_set`.

drnimbusrain:
@zmoon Thank you Zach.  What do you suggest to change these to?

zmoon:
Couldn't think of anything good for `pai_set`. For the `rk` I meant in the namelist reading subroutine change `real` to `real(rk)` ?

drnimbusrain:
@zmoon Can you check again on my usage of rk for pai_set?  You are correct that it had an issue with the way I had it before.  I remove the rk now for this.

zmoon:
I'll check it out later. Don't know why couldn't just use `real(rk)` for it everywhere.

drnimbusrain:
@zmoon OK, yes.  I make now all namelist real variables to rk consistently.  This should be better.

PR #35, opened 2022-11-10
Fixed midflameheight to use ceiling instead (test OK).

drnimbusrain:


zmoon:
does this imply the top flamelay could be only partial?

drnimbusrain:
@zmoon Can you further clarify?   I just wanted to move it to the upper layer if there is a partial flamelays.  Maybe the 2nd ceiling isn't necessary?

drnimbusrain:
@zmoon   FYI.  Wei-Ting just tested this new formulation, and the WAF results are looking really good against the Massman example figures in his paper.

zmoon:
I thought that originally you wanted to ensure that flamelays were all full-flame layers.

drnimbusrain:
@zmoon Ah, yes I see what you are saying considering my previous thoughts.  You are right, and I changed to

flamelays = floor(FLAMEH/MODRES) + 1
MIDFLAMEPOINT = max(ceiling(flamelays/2.0),2)

I change in latest [commit](https://github.com/noaa-oar-arl/canopy-app/pull/35/commits/93b014d0afeca17cf7b51803c58cdafe1363f34b), and this formulation seems to give best answer for midflameheight and layer.

drnimbusrain:
@zmoon Can we now merge this with Main?  Thanks!

zmoon (approved):

PR #39, opened 2023-01-23
Added new flameh_opt to use both FRP and User set value.

drnimbusrain:


drnimbusrain:
@angehung5 Since I have numerous more changes merged to develop now, I have decided for ease of your testing to merge in this new flameh_opt, since it is just an option.  Go ahead and pull down the latest develop and use that for testing.

PR #42, opened 2023-01-30
Fixed ny/nx = nlat/nlon bug in netcdf read.

drnimbusrain:


drnimbusrain:
@zmoon   Merging small NetCDF read bug into main without review.

PR #2, opened 2022-06-27
Update canopy_driver.F90

angehung5:
Add a function in canopy_driver.F90 to output a text file

PR #43, opened 2023-01-30
Fixed ny/nx = nlat/nlon bug in netcdf read.

drnimbusrain:


angehung5:
Model reads 2D netcdf input file successfully but gets following errors when writing netcdf output:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7fdd5d72787f in ???
#1  0x411bd1 in ???
#2  0x401c6a in ???
#3  0x7fdd5d7137b2 in ???
#4  0x401cad in ???
#5  0xffffffffffffffff in ???

Looks like model create an empty netcdf file but fails to write values

Issue #29, opened 2022-10-05
Add New Biogenic Emissions Feature


drnimbusrain:
Based on the LES chemistry simulations of [Clifton et al. (2022)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022MS003078), and after issue #21  - #28 are completed, we can add simplified submodules representing biogenic emissions, chemical reactions, dry deposition/[CRFsoilnox](https://github.com/geoschem/HEMCO/blob/main/src/Extensions/hcox_soilnox_mod.F90#L1985) as new features to the canopy-app.

drnimbusrain:
@zmoon Started initial [branch](https://github.com/noaa-oar-arl/canopy-app/tree/feature/bioemi) for a parameterized canopy biogenic isoprene emissions model for broadleaf leaves.  This is based on MEGANv3 and a combination of [Clifton et al. (2022)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022MS003078) and [Silva et al. (2020)](https://gmd.copernicus.org/articles/13/2569/2020/).    

It is not fully implemented yet, but initial tests create what seem to be a reasonable isoprene biogenic emissions profile (compared to Silva et al. 2020) for the canopy. Once issue #41 is addressed to enhance the example txt/ncf input files, I can fully integrate as a new option and test fully.  
 Z/Hc=   0.0000000000000000          ISOP_EMI (umol/m2 hr) =   1.9335550856135315E-002
 Z/Hc=   2.6315789473684209E-002 ISOP_EMI=   1.8967588519345848E-002
 Z/Hc=   5.2631578947368418E-002 ISOP_EMI=   1.8615665753298358E-002
 Z/Hc=   7.8947368421052627E-002 ISOP_EMI=   1.8280842367729203E-002
 Z/Hc=  0.10526315789473684      ISOP_EMI=   1.7964992408099862E-002
 Z/Hc=  0.13157894736842105      ISOP_EMI=   1.7671080710774414E-002
 Z/Hc=  0.15789473684210525      ISOP_EMI=   1.7403519451140052E-002
 Z/Hc=  0.18421052631578946      ISOP_EMI=   1.7168622302706556E-002
 Z/Hc=  0.21052631578947367      ISOP_EMI=   1.6975180671047349E-002
 Z/Hc=  0.23684210526315788      ISOP_EMI=   1.6835199830554117E-002
[...]

PR #50, opened 2023-02-16
[Develop] Feature/Biogenic Emissions

drnimbusrain:
This PR merges into develop and new in-canopy resolved biogenic emissions option (ifcanbio) for Canopy-App. It is based on MEGAN v2.1 and MEGANv3 and a combination of [Guenther et al. (2012)](https://gmd.copernicus.org/articles/5/1471/2012/), [Clifton et al. (2022)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022MS003078) and [Silva et al. (2020)](https://gmd.copernicus.org/articles/13/2569/2020/). There are a total of 19 biogenic emissions species based on Guenther et al. (2012), and the README has been updated to describe them.  This addresses in part issue #29. 

There are also numerous other updates to Canopy-App to facilitate improved output options given the large number of 3D biogenic species in this option. Moved to output NetCDF-4 format, and now the NL user canopy process flags (e.g., ifcanwind, ifcanwaf, ..., ifcanbio) control if the output will be written to netcdf file (when iofmt option is selected).

Output of all biogenic emissions are in kg m-2 s-1. Example Global and CONUS plots below:

image

image

Also note, since we have unified the inputs to be based on UFS type RRFS or GFS meteorology/land physics, I have added numerous more variables input to the model.    This is in preparation for further updates to Canopy-App in issue #29 for drydep, soilNOx CRFs, etc.  Right now, the only additional two variables the biogenic module needs are tmp2m and dswrf (instantaneous downward shortwave radiation).


README.md
```diff
@@ -27,62 +27,119 @@ Canopy is parameterized by foliage distribution shape functions and parameters f
...
+
+    **More Information on Data Sources from Table 2:**
+
+    **Downloading GFS Files from AWS:** NOAA's hourly global GFS, gridded (at ~13x13 km resolution) data may be downloaded publicly from the following Amazon Web Service (AWS) S3 location:
+    `https://nacc-in-the-cloud.s3.amazonaws.com/inputs/YYYMMDD/gfs.t12z.sfcfHHH.nc`  
+    Where HHH pertains to the hour of the 24-hr forecast (e.g., f000 is initialization).  
+    Example download command using wget: `wget --no-check-certificate --no-proxy https://nacc-in-the-cloud.s3.amazonaws.com/inputs/20230215/gfs.t12z.sfcf000.nc`
+    Hourly gridded GFSv16 data is available on AWS from March 23, 2021 - Current Day.
+
+    **GriddingMachine:** GriddingMachine is open source database and software for Earth system modeling at global and regional scales.  Data is easily accessible in consistent formats for ease of downloading/processing.  All avaialable datasets may be found at:  https://github.com/CliMA/GriddingMachine.jl. (Wang et al., 2022; https://doi.org/10.1038/s41597-022-01346-x)
```
zmoon:
```suggestion
    **GriddingMachine:** GriddingMachine is open source database and software for Earth system modeling at global and regional scales.  Data is easily accessible in consistent formats for ease of downloading/processing.  All available datasets may be found at:  https://github.com/CliMA/GriddingMachine.jl. (Wang et al., 2022; https://doi.org/10.1038/s41597-022-01346-x)
```

README.md
```diff
@@ -27,62 +27,119 @@ Canopy is parameterized by foliage distribution shape functions and parameters f
...
+
+    **Table 2. Canopy-App Required Input Variables**
+
+    | **GFS /Met/Land/Soil Variables**    | **Variable Description and Units**                           |  **Example Data Sources/References (if necessary)**                                                |
+    | ----------------------------------  | ------------------------------------------------------------ |  ------------------------------------------------- 						      |
+    | lat                                 | Latitude  (degrees)                                          |  N/A                                               						      |
+    | lon                                 | Longitude (degrees; from 0-360)                              |  N/A                                               					              |
+    | time                                | Timestamp (days since YYYY-N-D 0:0:0) (NetCDF Only)          |  N/A                                               					              |  
+    | ugrd10m                             | U wind at HREF (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16 *(see below for downloading using AWS)                                 						      |
```
angehung5, line 104: | ugrd10m                             | U wind at reference height above canopy (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16 *(see below for downloading using AWS)                                 						      |

README.md
```diff
@@ -27,62 +27,119 @@ Canopy is parameterized by foliage distribution shape functions and parameters f
...
+    | vgrd10m                             | V wind at HREF (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16                                   						      |
```
angehung5, line 105: | vgrd10m                             | V wind at reference height above canopy (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16                                   						      |

README.md
angehung5: | ugrd10m                             | U wind at reference height above canopy (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16 *(see below for downloading using AWS)                                 						      |

README.md
angehung5: | vgrd10m                             | V wind at reference height above canopy (m/s), e.g., 10 m                             |  UFS NOAA/GFSv16                                   						      |

README.md
```diff
@@ -27,62 +27,119 @@ Canopy is parameterized by foliage distribution shape functions and parameters f
...
+    | vtype                               | Vegetation type (dimensionless), VIIRS or MODIS              |  UFS NOAA/GFSv16                                   						      |
+    | fricv                               | Friction velocity (m/s)                                      |  UFS NOAA/GFSv16                                   						      |
+    | sfcr                                | Total surface roughness length (m)                           |  UFS NOAA/GFSv16                                   						      |
+    | frp              			  | Total Fire Radiative Power (MW/grid cell area)               |  UFS NOAA/GFSv16                                   						      |
```
angehung5: duplicate of line 125

README.md
drnimbusrain: @angehung5 Thank you.  I have modified due to your suggestions in recent commit.

README.md
drnimbusrain: @angehung5 Thank you.  I have modified due to your suggestions in recent commit.

README.md
drnimbusrain: @angehung5 Thank you.  I have modified due to your suggestions in recent commit.

drnimbusrain:
@zmoon @angehung5 Any more comments on this PR before merging?  It would be helpful if some tests could be run to evaluate the magnitude of the summed isoprene emissions for different months. On that note, @angehung5, how are the 12 monthly combined, global test files coming along? 

For example, comparing against the total surface MEGANv2.1 isoprene emissions in a recent Online-CMAQ run for July 22, 2022 at 12 UTC, the spatial pattern is very similar to our canopy biogenic isoprene emissions in eastern U.S.:

image

However, the magnitude is much smaller compared to our canopy sum isoprene emissions (note change in color scale by two orders of magnitude), and I am suspicious of my surrogate/approximate approach to calculating LAD in each canopy model layer, shown [here](https://github.com/noaa-oar-arl/canopy-app/blob/develop/canopy_bioemi_mod.F90#L191).

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,199 @@
...
+        GammaPPFD_SUN   = CP_SUN*((ALPHA_P_SUN*PPFD_SUN)/SQRT(1.0 + (ALPHA_P_SUN**2.0) * (PPFD_SUN**2.0)))
+        GammaPPFD_SHADE = CP_SHADE*((ALPHA_P_SHADE*PPFD_SHADE)/SQRT(1.0 + (ALPHA_P_SHADE**2.0) * (PPFD_SHADE**2.0)))
+
+        GammaPPFD_AVE = (GammaPPFD_SUN*RJCF) + (GammaPPFD_SHADE*(1.0-RJCF)) ! average = sum sun and shade weighted by sunlit fraction
+
+! Calculate isoprene emissions profile in the canopy
+        EMI_OUT = 0.0_rk  ! set initial emissions profile to zero
+        do i=1, SIZE(ZK)
+            if (ZK(i) .gt. 0.0 .and. ZK(i) .le. FCH) then  ! above ground level and at/below canopy top
+                FLAI(i) = ((FCLAI(i+1) - FCLAI(i)) * LAI)/FCH    !fractional LAI scaled to canopy depth (~ LAD)
```
zmoon: For layerwise LAD you would divide by the layer thickness (δz). LAD is dLAI/dz.

drnimbusrain:
Yes, definitely agreed.   I need to rework the  LAD likely.  Changed the LAD to dLAI/dz (MODRES).

zmoon:
In the top figure, is it per m2 grid cell area or per m2 vegetation?

canopy_ncf_io_mod.F90
```diff
@@ -340,80 +340,370 @@ SUBROUTINE canopy_outncf_init
...
+            c_rjcf%istart(3) = 1
+            c_rjcf%iend(1) = nlon
+            c_rjcf%iend(2) = nlat
+            c_rjcf%iend(3) = modlays
+        end if
+        if (ifcanbio) then
+            c_emi_isop%fld = fillreal
+            c_emi_isop%fldname = 'emi_isop'
+            c_emi_isop%long_name = 'biogenic isoprene emissions'
+            c_emi_isop%units = 'kg m2 s-1'
```
zmoon:
```suggestion
            c_emi_isop%units = 'kg m-2 s-1'
```
and the other ones too

drnimbusrain:
@zmoon Good point on the units.  It may in fact be per m2 vegetation considering on how I formulate the emission calculations.  This and the LAD approximate I need to check on.

canopy_bioemi_mod.F90
drnimbusrain: Yes, definitely agreed.   I need to rework the  LAD likely.

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,199 @@
...
+                EMI_OUT(i) = FLAI(i) * EF * GammaTLEAF_AVE(i) * GammaPPFD_AVE(i)  ! (ug/m2 hr)
+                EMI_OUT(i) = EMI_OUT(i) * 2.77778E-13 !TBD:  convert emissions output to kg/m2 s
```
zmoon:
```suggestion
                EMI_OUT(i) = FLAI(i) * EF * GammaTLEAF_AVE(i) * GammaPPFD_AVE(i)  ! (ug m-2 hr-1)
                EMI_OUT(i) = EMI_OUT(i) * 2.7777777777778E-13_rk !TBD:  convert emissions output to (kg m-2 s-1)
```
could add other `_rk`s as well in this module

drnimbusrain:
@zmoon I think I have fixed up the LAD calculation in latest commit, since indeed we are using Eq. (2) of Clifton et al. (2022) that has LAD (m2/m3):
image

The activity factors, gammaT1 and gammaPPFD, are unitless.  

The emission capacities/factors (EFs), epsilon, are in units of ug/m2 hr, and they are multiplied by LAD, in m2/m3.  This gives me units of ug/m3 hr, which you saw I convert to kg/m3 s  with the 2.8E-13 factor.  This is the amount of emissions per volume of the layer.  I think I need to [update the code](https://github.com/noaa-oar-arl/canopy-app/pull/50/commits/bd9dcfc9de434378a4f78346fefe32e8990b265c) to be reflective of this, and update my post analysis scripts to more accurately calculate the total summed emissions in kg/m2 s to compare against model surface MEGAN biogenics.   To me, I would then sum the emissions in the canopy, and multiply by canopy height to get the kg m-2 s-1 units, but this results in even larger biogenic emissions.

But I am not sold on the kg m-3 s-1 units necessarily as the output, and I look forward to your input here.

zmoon:
I kind of like the volume units for the vertical profile. If you integrate over the canopy it goes back to the flux (density) units, kg m-2 s-1, as you said. It is like $\sum_i{E_i \ \delta z_i}$, but so far $\delta z_i$ is constant so you can pull it out (i.e., multiply by $\delta z$ not $h_c$).

drnimbusrain:
@zmoon I agree, and will keep it as layer volume units. I have also updated the README to be consistent.

drnimbusrain:
@zmoon OK, I have re-analyzed the new LAD and volume units, and then summed the layers after converting to kg m-2 s-1 by multiplying by dz:

image

Still good spatial distribution of course, but the magnitude of summed emissions are at least 1 order of magnitude larger, and currently about 1000 times larger than comparison against surface MEGANv2.1 emissions in Online-CMAQ.

zmoon:
What about converting to per unit grid cell area?

drnimbusrain:
@zmoon What is your suggestion for that conversion?

zmoon:
Well if you had the fraction of the grid cell that is vegetation and the fraction of vegetation that is the plant type you computed the emission for, then you could multiply them.

drnimbusrain:
Is that what MEGAN is doing to get the units of kg m-2 s-1, as truly per "unit grid cell area"? If so, do you have where it does that?   Also, we do have the forest fraction for the grid cell (not vegetation fraction), but we do not have fraction of plant types within the dominant vtype for the grid cell.  For any given dominant vtype covering the entirety of grid cell, we simply take an average of the EFs for that grid cell from different mapped PFTs from Guenther et al. 2012.

zmoon:
"fractional grid box areal coverage", which I was getting at, is in the MEGAN2.1 paper eq. 1. I'll try to look for code later.

> we do not have fraction of plant types within the dominant vtype for the grid cell

What about vtype fraction out of all vtypes though?

> vtype covering the entirety of grid cell

If this is the case, that area fraction would be 1, so no conversion necessary.

drnimbusrain:
@zmoon  

Ah, now I am following what you are saying about the "fractional grid box area coverage" and Eq. 1 of Guenther et al. 2012.  No need for codes, and definitely in the future this biogenic module should use fractional vtype (i.e., landuse).     

However,  right now we are using dominant vtype (i.e., landuse type) as the input to Canopy-App biogenics that assumes the entirety of grid cell is covered by a single dominant vtype.  This is definitely a large simplification, even at ~13 km inputs. Of course inside the host model there is access to the fractional vtypes and this calculation can/should be done theoretically after Canopy-App to accurately use the correct emission fluxes in kg m-2 s-1.  

Still not sure if this is the real source of larger magnitude in Canopy-App vs. MEGANv2.1 output.  Even with fractional grid box areal coverage of say 51% for a given dominant vtype, the magnitude would still be much larger for Canopy-App vs. MEGAN2.1 example above.

drnimbusrain:
@zmoon Furthermore, to me it seems that dominant vtype (fraction = 1) is used in [MEGANv2.1 extension in HEMCO](https://github.com/geoschem/HEMCO/blob/main/src/Extensions/hcox_megan_mod.F90), which is used in the Online-CMAQ example above and produces much lower values.  I am wondering if it is my initial assumptions in my activity factors leading to much higher emissions.  I will revise and test.

README.md
```diff
@@ -46,7 +46,7 @@ Current Canopy-App components:
 
     - `canopy_phot_mod.F90`
 
-4.  In-Canopy vertical biogenic emissions (kg m-3 s-1). Note the units are in per m3 (layer volume) for vegetation type in grid cell/point.  Based on MEGANv2 and v3 (Guenther et al., 2012), and using both Clifton et al. (2021) and Silva et al. (2020) parameterizations.
+4.  In-Canopy vertical biogenic emissions (kg m-3 s-1). Note the units are in per m3 (layer volume) for vegetation type in grid cell/point. To get 2D total flux of biogenic emissions per m2 for each grid cell, the emissions profile would need to be integrated through the canopy layers, i.e., E*dz (or simply summed if dz is constant), and then multiplied by the fractional grid box areal coverage for each vegetation types.  For the current dominant vtype approach as input to Canopy-App, this multiplicative fraction = 1.  Based on MEGANv2 and v3 (Guenther et al., 2012), and using both Clifton et al. (2021) and Silva et al. (2020) parameterizations.
```
zmoon: $\sum_i E_i \delta z_i = \delta z \sum_i E_i$ if $\delta z_i = \delta z$ (constant), i.e. still multiplying dz, but you can do it after the sum if dz is constant.

drnimbusrain:
@zmoon I fixed my issue in linear interpolation of Tleaf from T2 based on Silva et al., and I also added the needed "Canopy Environment Coefficient (CCE)" used in MEGAN and from Guenther et al. (2012) and described as a tuning factor (e.g., in Clifton et al. 2022) needed to get reasonable emission factors in the canopy from MEGAN default values.   Right now I base it on the derived  value of 0.21 from Silva et al. (2020), but of course it is dependent on the model system and inputs used to get the starndard conditions.  I will next likely make this a NL option in order to play with value for best "tuning".  After these changes, I am getting more reasonable magnitudes, which are likely still high because of the previously discussed issues with using dominant vtype as input across relatively coarse resolution and thus not really adjusting (factor = 1) for the associated vtype "fractional grid box area coverage".  

image


However, our remaining larger summed ISOP emission magnitudes compared to our Online-CMAQ model that uses [HEMCO/MEGANv2.1 in NEXUS](https://github.com/geoschem/HEMCO/blob/main/src/Extensions/hcox_megan_mod.F90) with [GFSv16 input data (that looks like MERRA2)](https://github.com/noaa-oar-arl/NEXUS/blob/develop/config/cmaq/NEXUS_Config.rc#L1646) is still interesting because this is also based on dominant vtype for the grid cell with no information of fractional vtypes (land use type) in its calculations.

README.md
drnimbusrain: @zmoon I have modified the README.

drnimbusrain:
@zmoon  I think Barry raised an interesting point at the meeting today.  Do you think the main issue is that we should only sum certain canopy layers to get the "surface flux" emissions, if so, how many layers do you think?  Maybe more insight into how MEGAN generates that flux based on big-leaf with some "canopy layer effects" taken into account (esp. MEGANv3).  I also agree with you that we are generating canopy layer volume based leaf-level emissions, and is not taking into account the vertical transport.

drnimbusrain:
@zmoon Also, had another thought that in Canopy-App case its not necessarily about summing them through the canopy layers, but taking into eddy diffusivity transport in the canopy.  Just trying to think how to relate the resolved canopy layer emissions to a 2D flux at top of canopy to compare to our MEGAN predictions. 

I think the main point is that neither MEGANv2 or MEGANv3 truly calculate the emissions at leaf level (with LAD) as we are doing here in Canopy-App.

zmoon:
> only sum certain canopy layers to get the "surface flux" emissions

Assuming no chemical/deposition/etc. loss, all of the leaf-level layerwise emissions *will* escape the canopy, it's just a matter of timing. That's what modeling the vertical turbulent transport would be able to give us, but we're not currently doing time integration stuff. So we do want to integrate over the whole canopy in some way to get the overall canopy flux, incorporating the total mass that the canopy has produced, which scales with total LAI.

> emissions at leaf level (with LAD)

I know one of the MEGAN codes (maybe v3) I have looked at did sum emissions from different canopy layers to get the total. I'll try to find some other examples.

drnimbusrain:
thanks @zmoon OK, yes, I see what your saying, but assuming that MEGAN emissions are not doing any chemical/deposition loss scaling, and we likely then are overestimating the summed emissions from Canopy-App compared to MEGAN, unless MEGAN is doing something to reduce emissions based on losses from layers scaled to total canopy (LAI):

image
image


Overall, the profile looks reasonable with LAD, but just curious on the magnitude:
  
image

drnimbusrain:
@zmoon Definitely agree to check MEGANv3 codes again to see how they may sum canopy emissions, but to me it seems they are using activity factors calculated jointly per canopy level and summed together _weighted by the vertical canopy biomass_ _distribution_, and then multiply later by the total effect of LAI. 

I think this is where MEGAN3 calculates and sums the activity response to canopy temperature/light

        IF ( Layers .EQ. 5 ) THEN
          VPGWT(1) = 0.1184635
          VPGWT(2) = 0.2393144
          VPGWT(3) = 0.284444444
          VPGWT(4) = 0.2393144
          VPGWT(5) = 0.1184635
        ELSE
          DO K = 1,Layers
          VPGWT(K) =1/Layers
          ENDDO
        ENDIF

          DO I = 1, NCOLS
            DO J = 1, NROWS

              DO K = 1, Layers
                Ea1L(K)  = CDEA(I,J,K) *
     &                GAMTLD(SunT(I,J,K),D_TEMP(I,J),S) *
     &                GAMP(SunP(I,J,K),D_PPFD(I,J)) *  SunF(I,J,K) +
     &                GAMTLD(ShaT(I,J,K),D_TEMP(I,J),S) *
     &                GAMP(ShaP(I,J,K),D_PPFD(I,J))
     &                * (1-SunF(I,J,K))

                Ea2L(K) = GAMTLI(SunT(I,J,K),S)* SunF(I,J,K)+
     &              GAMTLI(ShaT(I,J,K),S)*(1-SunF(I,J,K))

              ENDDO   ! ENDDO canopy layers
            GAMTP(I,J)=SUM((Ea1L(:)*LDFMAP(I,J) +
     &                      Ea2L(:)*(1-LDFMAP(I,J)))* VPGWT(:))
            ENDDO   ! NROWS
          ENDDO ! NCOLS




Maybe we do something similar, but we are calculating the actual leaf-level emissions at LAD.  Overall, I wonder if I have another issue in my code with units etc...

drnimbusrain:
@zmoon I have found a few errors in my PPFD (PAR) calculation through the canopy and a bug.  I think my latest commit now fixes the issue of Canopy-App being much too large after summing the layers and comparing to Online-CMAQ MEGANv2.1 output, both updated plots on the same scale in kg m-2 s-1 below:

image
image
We are have good agreement in spatial pattern and now the same order of magnitude compared to Online-CMAQ MEGANv2.1 (max ~ 3.9 times higher for Canopy-App).

This is a direct sum in Canopy-App multiplied by the constant layer resolution (0.5 m), and I do not weight the layers by biomass distribution as MEGAN does when summing (above).  I think we have already calculated the leaf-level emissions based on biomass distribution profile, LAD.  So, I think this distribution weighting effect is already accounted for in Canopy-App.  Let me know what you think.

drnimbusrain:
And the biogenic emissions profile looks much better now:

image

zmoon:
> This is a direct sum in Canopy-App multiplied by the constant layer resolution (0.5 m), and I do not weight the layers by biomass distribution as MEGAN does when summing (above). I think we have already calculated the leaf-level emissions based on biomass distribution profile, LAD. So, I think this distribution weighting effect is already accounted for in Canopy-App. Let me know what you think.

Yes we are already effectively doing this weighting by using the LAD profile to scale the emissions at given layer.

drnimbusrain:
@zmoon Great, then it seems to me that we have a reasonable comparison now for Canopy-App biogenics!

canopy_alloc.F90
```diff
@@ -70,5 +70,51 @@ SUBROUTINE canopy_alloc
...
     end if
 
+!-------------------------------------------------------------------------------
+! Allocate arrays for Canopy Biogenic Emissions Outputs
+!-------------------------------------------------------------------------------
+
+    if (ifcanbio) then
+        write(*,*)  'Canopy biogenic emissions option selected'
+        write(*,*)  '-------------------------------'
+        if(.not.allocated(emi_isop))         allocate(emi_isop(nlat*nlon,modlays))
```
zmoon, line 80: I think a lot of the code could be made more extendable and maintainable if we used internally a single 4-d emi array instead, with species dimension.

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+        REAL(RK),    INTENT( IN )       :: TEMP2           ! Model input 2-m Temperature (K)
+        INTEGER,     INTENT( IN )       :: LU_OPT          ! integer for LU type from model mapped to Massman et al. (default = 0/VIIRS)
+        INTEGER,     INTENT( IN )       :: VTYPE           ! Grid cell dominant vegetation type
+        REAL(RK),    INTENT( IN )       :: MODRES          ! Canopy model input vertical resolution (m)
+        REAL(RK),    INTENT( IN )       :: CCE             ! MEGAN Canopy environment coefficient.
+        INTEGER,     INTENT( IN )       :: EMI_IND         ! Input biogenic emissions index
+        REAL(RK),    INTENT( OUT )      :: EMI_OUT(:)      ! Output canopy layer volume emissions (kg m-3 s-1)
+
+! Local Variables
+        REAL(RK) :: RJCF(SIZE(ZK))                 ! Photolysis correction factor for sun/shade fraction of leaf layer
```
zmoon: I would call this something that emphasizes sunlit leaf fraction, perhaps `fsun`, since that is how it is being used.

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+        REAL(RK),          PARAMETER     :: DTEMP_2_SHADE   =  -0.141_rk  !...
+        REAL(RK),          PARAMETER     :: DTEMP_3_SHADE   =  -0.368_rk  !...
+        REAL(RK),          PARAMETER     :: DTEMP_4_SHADE   =  -0.592_rk  !...
+        REAL(RK),          PARAMETER     :: DTEMP_5_SHADE   =  -0.743_rk  !...
+
+        REAL(RK),          PARAMETER     :: CT2             =  230.0_rk   !Deactivation energy (kJ/mol) (Guenther et al., 2012)
+
+!Calculate photolyis shading/correction factor through canopy, i.e., the fraction of sunlit leaves downward through canopy
+
+        call canopy_phot(FCLAI, LAI, CLU, COSZEN, RJCF)
```
zmoon: In the clumping case, relative direct beam irradiance (our current photolysis attenuation factor) is not equivalent to sunlit leaf fraction. See Bonan eq. 14.18 (need to multiply by clumping factor).

Without clumping, it would be. But in general they aren't necessarily the same (e.g. with different radiative transfer approach).

canopy_bioparm_mod.F90
```diff
@@ -0,0 +1,802 @@
...
+        REAL(RK),    INTENT( OUT )      :: EF              ! Out Mapped EF ((ug/m2 hr)
+        REAL(RK),    INTENT( OUT )      :: CT1             ! Out Activation energy (kJ/mol)
+        REAL(RK),    INTENT( OUT )      :: CEO             ! Out Empirical coefficient
+!  LOCAL
+        REAL(RK) :: EF1,EF2,EF3,EF4,EF5,EF6,EF7    ! Plant Emission factors (EF) (ug/m2 hr)
+        REAL(RK) :: EF8,EF9,EF10,EF11,EF12,EF13    !
+        REAL(RK) :: EF14,EF15                      !
+
+! Plant-Dependent emissions capacity/factors (EFs) for Isoprene (Tables 2-3 of Guenther et al., 2012) (ug/m2 hr)
+        REAL(RK),          PARAMETER     :: EF1_ISOP    =  600.0_rk     ! Needleleaf Evergreen Temperate Tree
```
zmoon, line 50: could place the EFs in a single 2-d array with species dimension

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+                DTEMP_SHADE(i) = 0.0
+            else if (ZK(i) .le. FCH .and. ZK(i) .gt. FCH*(4.0_rk/5.0_rk)) then  !Level 1 - 2
+                CTEMP_SUN(i)   = interp_linear1_internal((/ FCH*(4.0_rk/5.0_rk),FCH /), &
+                    (/ CTEMP_2_SUN,CTEMP_1_SUN /),ZK(i))
+                DTEMP_SUN(i)   = interp_linear1_internal((/ FCH*(4.0_rk/5.0_rk),FCH /), &
+                    (/ DTEMP_2_SUN,DTEMP_1_SUN /),ZK(i))
+                CTEMP_SHADE(i) = interp_linear1_internal((/ FCH*(4.0_rk/5.0_rk),FCH /), &
+                    (/ CTEMP_2_SHADE,CTEMP_1_SHADE /),ZK(i))
+                DTEMP_SHADE(i) = interp_linear1_internal((/ FCH*(4.0_rk/5.0_rk),FCH /), &
+                    (/ DTEMP_2_SHADE,DTEMP_1_SHADE /),ZK(i))
```
zmoon, line 231: this part is a bit ugly, could instead set some local variables, for what to interp and then do the interp outside the loop

or refactor so that the coeffs are in arrays instead of separate variables

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+!physics surrogate model for use in chemical transport models: a case study with GEOS-Chem v12.3.0,
+!Geosci. Model Dev., 13, 2569–2585, https://doi.org/10.5194/gmd-13-2569-2020, 2020.
+        do i=1, SIZE(ZK)  !calculate linear change in parameters interpolated to Silva et al. 5 layer canopy regions
+            if (ZK(i) .gt. FCH) then ! above canopy, Tleaf = Tair
+                ATEMP_SUN(i)   = 0.0
+                BTEMP_SUN(i)   = 1.0
+                ATEMP_SHADE(i) = 0.0
+                BTEMP_SHADE(i) = 1.0
+            else if (ZK(i) .le. FCH .and. ZK(i) .gt. FCH*(4.0_rk/5.0_rk)) then  !Level 1 - 2
+                ATEMP_SUN(i)   = interp_linear1_internal((/ FCH*(4.0_rk/5.0_rk),FCH /), &
```
zmoon, line 169: same comment as below

namelist.canopy
```diff
@@ -1,7 +1,7 @@
 &FILENAMES
   file_vars    = 'gfs.t12z.20220701.sfcf000.nc'
 !  file_vars    = 'gfs.t12z.20220701.sfcf000.txt'
-  file_out     = 'test_out'
+  file_out     = 'test_out_se'
```
zmoon: 'se'?

zmoon (commented):
Some initial comments.

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+        REAL(RK) :: E_OPT(SIZE(ZK))                ! maximum normalized emission capacity
+        REAL(RK) :: TLEAF_OPT(SIZE(ZK))            ! Tleaf at which E_OPT occurs (K)
+        REAL(RK) :: FLAI(SIZE(ZK))                 ! Fractional LAI in layer
+        REAL(RK) :: CT1                            ! Activation energy (kJ/mol)
+        REAL(RK) :: CEO                            ! Empirical coefficient
+        REAL(RK) :: EF                             ! Final Mapped Emission factor (EF) (ug/m2 hr)
+        integer i
+
+! Constant Canopy Parameters
+        REAL(RK),          PARAMETER     :: FRAC_PPFD       =  0.5_rk     !Fraction of incoming solar irradiance that is active PPFD
```
zmoon: "active PPFD"?? do you reference for this?

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+                DTEMP_SUN(i)   = interp_linear1_internal((/ ZK(1),FCH*(2.0_rk/5.0_rk) /), &
+                    (/ DTEMP_5_SUN,DTEMP_4_SUN /),ZK(i))
+                CTEMP_SHADE(i) = interp_linear1_internal((/ ZK(1),FCH*(2.0_rk/5.0_rk) /), &
+                    (/ CTEMP_5_SHADE,CTEMP_4_SHADE /),ZK(i))
+                DTEMP_SHADE(i) = interp_linear1_internal((/ ZK(1),FCH*(2.0_rk/5.0_rk) /), &
+                    (/ DTEMP_5_SHADE,DTEMP_4_SHADE /),ZK(i))
+            end if
+        end do
+
+        PPFD_SUN     = FRAC_PPFD * SFCRAD * EXP(CTEMP_SUN + DTEMP_SUN * LAI)  !W/m2 --> umol m-2 s-1
```
zmoon: 0.5 does not do the conversion to umol m-2 s-1. That's what the 4.5 was doing. 1 W m-2 ≈ 4.6 umol m-2 s-1 for the visible waveband for typical clear sky conditions. So as it stands I am a bit confused about this change.

zmoon (commented):
Some comments on the PPFD-ness.

canopy_alloc.F90
drnimbusrain: @zmoon I had thought about that, and tend to agree with you.  I thought even the species desired could be controlled though the NL, but didn't want to go down that path right away.  I wanted just something to work initially.

So, can you make a new branch off of this and start to make these changes for a 4-d "emi_bio" array instead,  and then we can compare how it looks?

canopy_bioemi_mod.F90
drnimbusrain: @zmoon Totally agree. Making the [change](https://github.com/noaa-oar-arl/canopy-app/pull/50/commits/57ae1f5da80b117e243e8bc6a3053e3d4de6d6e9).

canopy_bioemi_mod.F90
drnimbusrain: @zmoon Thank you for that.  We do get very reasonable sunlit/shaded fractions that when I test look very close to  this from Clifton et al.:

image

But, if not fully correct, please go ahead and make proposed changes and do a commit so we can test.  Seems also to produce reasonable profiles of Tleaf and PPFD for sunlit and shaded leaves too, as you show below.

canopy_bioparm_mod.F90
drnimbusrain: @zmoon   Yes, seems more compact and cleaner as your initial comment on the 4-d emissions array.  Can you go ahead and add these 2-d EF changes to the new branch with the 4-d emi_bio array as well?

canopy_bioemi_mod.F90
drnimbusrain: @zmoon I tend to agree with you, but gets the job done.  Please go ahead and make suggested changes and we can commit and test.  Thanks!

canopy_bioemi_mod.F90
drnimbusrain: @zmoon Same response as above...  ;)

canopy_bioemi_mod.F90
drnimbusrain: @zmoon   Yes, I understand.  The FRAC_PPFD is 50% the incoming total solar radiation, since the incoming SWRAD is not equal to the incident PPFD at top of canopy, as described above.  

The conversion seems to come the exponential relationship in equation (2) in Silva et al.

image

Seems here the input to the derived relationship is PARtoc in W/m2, which gets converted to the output in umol/m2 s.    I definitely was a bit confused as to what Silva was doing here, but thought the coefficients and exponential relationship derived had the conversion embedded by what is stated.  I agree the 4.5 I had there before could be added back, but need clarification.

canopy_bioemi_mod.F90
drnimbusrain: Yes, the PAR incident at the top of the canopy is not equivalent to the total instantaneous downward shortwave flux (SWRAD) from the model, but can approximated as a first guess here at 50% the SWRAD.  I thought PAR was in the visible spectrum at about 400-700 nm, which in essence must then be a fraction of the SWRAD that is in the broader shortwave spectrum from about 300 - 4000 nm.  Definitely open to what you may think this fraction should be though...

zmoon: Should call it `frac_par` instead (because of the difference in units, the fractions could be different). I think 0.5 sounds fine but I will check...

Or another option would be to make PAR the input, the fraction already applied.

canopy_bioemi_mod.F90
zmoon: Yeah does seem to suggest that the coeffs somehow include the conversion. I may briefly try it out.

namelist.canopy
drnimbusrain: This just refers to southeast us example.  Will update it to be clearer.  See latest [commit](https://github.com/noaa-oar-arl/canopy-app/pull/50/commits/15a32e1318665030f237f9401ae023036387daa8).

canopy_bioemi_mod.F90
zmoon: <details>
<summary>Seems to check out</summary>

silva-coeffs_fig1
</details>

canopy_bioemi_mod.F90
```diff
@@ -0,0 +1,310 @@
...
+!Silva, S. J., Heald, C. L., and Guenther, A. B.: Development of a reduced-complexity plant canopy
+!physics surrogate model for use in chemical transport models: a case study with GEOS-Chem v12.3.0,
+!Geosci. Model Dev., 13, 2569–2585, https://doi.org/10.5194/gmd-13-2569-2020, 2020.
+        do i=1, SIZE(ZK)  !calculate linear change in parameters interpolated to Silva et al. 5 layer canopy regions
+            if (ZK(i) .gt. FCH) then ! above canopy, Tleaf = Tair
+                ATEMP_SUN(i)   = 0.0
+                BTEMP_SUN(i)   = 1.0
+                ATEMP_SHADE(i) = 0.0
+                BTEMP_SHADE(i) = 1.0
+            else if (ZK(i) .le. FCH .and. ZK(i) .gt. FCH*(4.0_rk/5.0_rk)) then  !Level 1 - 2
```
zmoon, line 168: where did you get these boundary definitions? I don't see that in the paper

canopy_bioemi_mod.F90
drnimbusrain: @zmoon Yes I did an independent check and that is what I saw too.  I think it looks good, but I am open to using other methods maybe later to get at the fraction of sunlit/shaded leaves through the canopy.

canopy_bioemi_mod.F90
drnimbusrain: @zmoon They are not in the paper. I set these myself because in the linear Tleaf and exponential PPFDleaf equations using these boundary conditions gives back the T2 and PPFD above canopy (as expected).  

Tleaf = Tair (when A = 1 and B = 0)
PPFD = PPFDtoc (when C=0 and D=0)

canopy_bioemi_mod.F90
drnimbusrain: @zmoon   IOK, [changed to FRAC_PAR](https://github.com/noaa-oar-arl/canopy-app/pull/50/commits/141d392bc2f9ae6bf5a203da5902c5975551e495).

canopy_bioemi_mod.F90
zmoon: I am going to do some more checks, e.g. comparing to my own calculation. One issue is that the relative plant area profile it is based on is fixed, and most likely not consistent with canopy-app's. But we could improve that later.

canopy_bioemi_mod.F90
zmoon: Sure, I meant the z boundaries. The $4/5 h_c$ and so on.

canopy_bioemi_mod.F90
drnimbusrain: @zmoon I agree this is an issue with fixing the Tleaf and PPFDleaf on these linear and exponential relationships from Silva et al.  Looking forward to what you find out to help improve this.

canopy_bioemi_mod.F90
drnimbusrain: @zmoon Ah sorry. Yes, these are just rough boundaries that break down the canopy depth into four sections and interpolate the regression coefficients A, B, C, and D between the 5 levels in Table  1 of Silva et al.  Again, not optimal and can be improved.  It does create reasonable Tleaf and PPFD through the canopy, but as mentioned above, the profile here as a relatively fixed shape, and not fully consistent with our Canopy-App LAD profiles varying with vegtype.  

image
image

Essentially this was a shortcut to avoid the leaf energy balance model approach in solving for Tleaf etc.  If we want to go this route, I don't think this would not be too hard to parameterize starting with MEGANv3 (e.g., MEGCAN on Hopper:  _/groups/ESS/pcampbe8/MEGAN/MEGANv3.2_Dec_2021/src/MEGCAN/megcan.f_), and fully accounting for energy fluxes, humidity, wind speed etc. through the canopy.  

Do you want to give it a shot?

zmoon: > Do you want to give it a shot?

Yeah probably later, for this PR let's focus on what you have already.

drnimbusrain: @zmoon   Sounds good, but I agree we should bring in some aspect of canopy layer and ground energy balance models into Canopy-App following other applications of MEGAN, Bonan's many works, or other applications of canopy parameterizations in gridded models, like [WRF MCANOPY](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2018MS001347).  This will allow us to do more sophisticated parameterizations of biogenics, deposition, soil fluxes, and things like biogeochemical fluxes of carbon etc.

drnimbusrain:
@zmoon   Not sure why the CI [wfindent executable](https://github.com/noaa-oar-arl/canopy-app/actions/runs/4471037165/jobs/7855384827?pr=50#step:4:112) is not working anymore...

zmoon:
> @zmoon Not sure why the CI [wfindent executable](https://github.com/noaa-oar-arl/canopy-app/actions/runs/4471037165/jobs/7855384827?pr=50#step:4:112) is not working anymore...

Sorry about that, fixed it for now, there was a change in the upstream `findent` PyPI package, not including `wfindent` anymore.

zmoon:
Currently we have a check that the default input is nc (like the build setting), not txt, that's why CI is failing. Maybe change it back?

canopy_alloc.F90
zmoon: @drnimbusrain I will try to work on this one soon.

canopy_bioparm_mod.F90
zmoon: Think will hold off on this, we can always come back to it.

canopy_bioemi_mod.F90
zmoon: Since this is internal, think I will hold off on this, we can always come back to it.

canopy_bioemi_mod.F90
zmoon: We will come back to this for sure.

drnimbusrain:
@zmoon Sorry I was just testing the new file names in default namelist, and forgot to switch back to default nc, which I think we should keep for now.

canopy_alloc.F90
drnimbusrain: @zmoon Thanks!  Makes sense to work on it here, but let me know if we want to merge this large PR first and then branch off, its up to you.

zmoon:
@drnimbusrain also need to change this line here

https://github.com/noaa-oar-arl/canopy-app/blob/33e4e2b703c5563bfffa9479d30161ad87ec5ef8/.github/workflows/ci.yml#L70

since you changed the file name, sorry for the inconvenience

canopy_alloc.F90
zmoon: > we want to merge this large PR first and then branch off, its up to you.

Yeahhh I think this is a better idea. We can come back to it if we want add to add a bunch more species at some point or something.

canopy_bioemi_mod.F90
zmoon: I think this is fine for now. We can come back to this at some point, maybe make the frac an input.

canopy_bioemi_mod.F90
zmoon: @drnimbusrain I updated in d065c28

canopy_bioemi_mod.F90
zmoon: <details><summary>A little example</summary>

image
</details>

Similar enough, though a little suspicious that the Silva profiles don't get very close to 4.6 (should be at L=0 for clear sky conditions). But let's leave it for now.

canopy_bioemi_mod.F90
drnimbusrain: @zmoon I did run a quick check and the isoprene emissions are rather consistent:  
image

I'd say this is a good change to be consistent approach with Bonan! Nice work.

zmoon (approved):
Great start to biogenic emissions

PR #57, opened 2023-04-13
Modified canopy wind calculation to be continuous.

### LINKED ISSUE #56: Fix Canopy Wind Speed Discontinuity with Massman/MOST Approach
@zmoon  pointed out the unphysical jump in the modeled wind profile from above the canopy to the canopy top (panel on far left below):

image

The main issue is that the canopy-wind codes were originally only designed to go from canopy-top downward, and then it got a bit fragmented when combining this approach with the MOST and unified RSL approaches I adopted from Noah-MP and Rosenzweig et al., (2021)  https://doi.org/10.1029/2021MS002665.

This could be rectified by making the wind profile smooth below the reference height wind (e.g., @ 10 m winds from model).  Reference height wind is needed in WAF calculation, so winds can be constant above the reference height wind.

Ultimately, we should have a smoothly varying wind speed below the reference height wind (such as in Bonan's plot below):

image

drnimbusrain:
This addresses issue #56 , and instead of setting winds to reference height at z > hc, this uses the log wind profile above canopy top.  This also removes the fragmented RSL_OPT (=1) effects option that created even more discontinuity across the canopy top from the different approaches.  Thus, RSL_OPT = 0 is the only option for now, and uses the LAMBDARS scaling parameter to represent RSL effects above the canopy top for now.

After the change the wind speed profile is more continuous, with still a relatively sharp change at the canopy top from logarithmic to exponential in type.  
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ https://nacc-in-the-cloud.s3.amazonaws.com/inputs/geo-files/canopy.2022MM01.test
 | `href_opt`      | integer for using href_set in namelist (= `0`, default) or array from file (= `1`) |
 | `href_set`      | user-set real value of reference height above canopy associated with input wind speed (m) (only used if `href_opt=0`) ***\*\*** |
 | `z0ghc`         | ratio of ground roughness length to canopy top height (Massman et al., 2017)       |
-| `rsl_opt`       | user-set option for Roughness SubLayer (RSL) effects above and at canopy top (Uc). `0`: uses a constant lambdars factor only |
+| `rsl_opt`       | user-set option for Roughness SubLayer (RSL) effects above and at canopy top (Uc).(= `0`, default: uses a constant lambdars factor only), (= `1`, default: uses a unified RSL approach from Bonan et al. (2018) and Abolafia-Rosenzweig et al., 2021)   |
 | `lambdars`      | Value representing influence of RSL effects (with `rsl_opt=0`) (Massman et al., 2017)          |
 | `dx_opt`        | `0`: Calculation of dx resolution/distance from lon; `1`: user-set dx grid resolution |
 | `dx_set`        | user-set real value of grid resolution (m) only if `dx_opt=1`                      |
@@ -212,6 +212,10 @@ https://nacc-in-the-cloud.s3.amazonaws.com/inputs/geo-files/canopy.2022MM01.test
 ## References
 
 *Further references contained within the source code.*
+
+- Abolafia-Rosenzweig, R., He, C., Burns, S. P., & Chen, F. (2021). Implementation and evaluation of a unified turbulence parameterization throughout the canopy and roughness sublayer in Noah-MP snow simulations. Journal of Advances in Modeling Earth Systems, 13, e2021MS002665. https://doi.org/10.1029/2021MS002665.
+
+- Bonan, G. B., Patton, E. G., Harman, I. N., Oleson, K. W., Finnigan, J. J., Lu, Y., & Burakowski, E. A. (2018). Modeling canopy-induced turbulence in the Earth system: A unified parameterization of turbulent exchange within plant canopies and the roughness sublayer (CLM-ml v0). Geoscientific Model Development, 11, 1467–1496. https://doi.org/10.5194/gmd-11-1467-2018.
 
 - Clifton, O. E., Patton, E. G., Wang, S., Barth, M., Orlando, J., & Schwantes, R. H. (2022). Large eddy simulation for investigating coupled forest canopy and turbulence influences on atmospheric chemistry. Journal of Advances in Modeling Earth Systems, 14, e2022MS003078. https://doi.org/10.1029/2022MS003078
 
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ https://nacc-in-the-cloud.s3.amazonaws.com/inputs/geo-files/canopy.2022MM01.test
 | `ifcanphot`     | logical canopy photolysis option (default: `.FALSE.`)                              |
 | `ifcanbio`      | logical canopy biogenic emissions option (default: `.FALSE.`)                      |
 | `href_opt`      | integer for using href_set in namelist (= `0`, default) or array from file (= `1`) |
-| `href_set`      | user-set real value of reference height above canopy associated with input wind speed (m) (only used if `href_opt=0`); Note that if href becomes small and approaches z0 (or as href --> 0), only the sub-canopy wind profile is calculated, recommend href = 10m  |
+| `href_set`      | user-set real value of reference height above canopy associated with input wind speed (m) (only used if `href_opt=0`) **\*\*** |
 | `z0ghc`         | ratio of ground roughness length to canopy top height (Massman et al., 2017)       |
 | `rsl_opt`       | user-set option for Roughness SubLayer (RSL) effects above and at canopy top (Uc). `0`: uses a constant lambdars factor only |
 | `lambdars`      | Value representing influence of RSL effects (with `rsl_opt=0`) (Massman et al., 2017)          |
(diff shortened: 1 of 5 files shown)


drnimbusrain:
@angehung5 You also may want to test this change with the canopy wind evaluation, as it may have some slight impacts on the results, particularly since RSL_OPT = 0 is the only option now to make the profile continuous in output.  This should have no effect on WAF calculation.

drnimbusrain:
@zmoon and @angehung5 Converted to draft, as this needs more thought...I am also unsure on how much the continuous profile is needed here in canopy-app, considering its application to gridded ATM/LSM models that already have computed the log MOST/RSL profile above the canopy...  Up for more discussion on approaches for sure.

angehung5:
Screen Shot 2023-04-13 at 1 38 34 PM

Figure shows the wind profile at Chestnut Ridge site. Updated code creates continuous profile. But I am somehow uncomfortable with removing RSL_opt =1, since it actually provides better results than RSL_opt=0 when reference height larger than 0.

drnimbusrain:
@angehung5 Thank you for your testing, and we could certainly find  a way to work back in RSL =1, but keep profile continuous when HREF > 0.   We can also tune the LAMBDARS in RSL=0 though.

angehung5:
Screen Shot 2023-04-14 at 3 54 04 PM
Just found a bug in my script. Here is the updated plot of wind profiles at Chestnut using updated code. A relative sharp change at canopy top like Patrick shown above.

And here shows the wind profiles using updated code with RSL_opt=0, old code with RSL_opt=0 and old code with RSL_opt=1. Seems like updated code tends to create higher wind speed at canopy top. @drnimbusrain How do you get the  log profile above canopy? Based on Massman?
Screen Shot 2023-04-14 at 4 13 09 PM

drnimbusrain:
There long profile is based on the Noah-MP calc (with RSL LAMBDARS factor),
but get very similar results when using the Massman log wind profile.  You
could try another test with updated RSL_OPT =0 code and lambdars = 1.0.
LAMBDARS is essentially a RSL tuning factor (with no representation of
atmospheric stability).  I will try to bring back the RSL_opt = 1 soon.

angehung5:
Great, thanks! I test lambdars = 1.0 with updated RSP_opt=0 code. It looks better now but the wind speed at canopy top is still higher than the old results.

drnimbusrain:
@angehung5 I have made some substantial changes (may still need a bit more work on an updated re-calculation of zero plane displacement height for RSL) to bring back an improved method for the unified RSL_OPT=1.  Can you please test it again with the above case for new RSL_OPT=0 and =1?  Thank you!

drnimbusrain:
@angehung5 I just tested these mods for continuous RSL_OPT=1, and it fails for regional domains in some locations.  I think you could still test it at Chestnut Ridge, but, we should discuss more about this maybe in person at the AMS meeting this week.  It is not trivial to bring back RSL_OPT=1 in  a full unified approach with continuous profile, energy balance, and iterative updates to the MOL for a canopy.  This is also complicated with coupling this approach (from Bonan's work) above the canopy to Massman within the canopy.

drnimbusrain:
@angehung5 @zmoon OK, I have made the canopy_wind module to have only the Massman/MOST subroutine (RSL_OPT=0), controlled through the canopy_calcs now.  The profile should be continuous below and above the canopy, with Massman below, and general MOST theory above.  Remember in Massman, RSL_Opt=0 approach, the RSL effects are simply the LAMBDARS variable than just provides a constant factor to tune the RSL effects.  I think this is sitll a valuable approach.

So, as noted in the README, the RSL_OPT=0 is the only option right now.

Once we can can take Bonan's codes from Zach, we can create another unified RSL approach (somehow) as a new subroutine in the canopy_wind module.  This can then form a better continuous, and unified RSL_OPT=1 approach, which will differ from the pure Massman/MOST approach.

What do you think?  Maybe if Wei-Ting can re-evaluate this new continuous Massman/MOST (RSL_OPT=0), I can clean up the code and we can merge, and then make a new issue/PR for integrating a more robust RSL_OPT=1 for unified RSL from Bonan?

angehung5:
Yes, I think it is a good idea to re-evaluate the new profile with observation. Not all sites have above-canopy wind measurements though. 

What do you mean merge? Merge to the develop branch?

drnimbusrain:
@angehung5 Yes, merge the continuous Massman/MOST profile (RSL_OPT=0 only) to develop branch so we can at least have the wind profile continuous using MOST for now, and then open a new issue/PR for the Bonan RSL approach.

drnimbusrain:
@angehung5 Please let me know when you have re-evaluated the new RSL_OPT=0 option here in 'canwind' branch, and if it looks OK through canopy. 

@zmoon Then we can finalize review of this branch, merge with develop branch, and then I can start to work on a new branch PR for RSL_OPT=1 branch with Bonan unified RSL approach ( issue #62 ) .

drnimbusrain:
@angehung5 When you have the updated global, gfs canopy files with all sfc meteorology variables, you can compare to those I made at:  _/groups/ESS3/pcampbe8/gfsv16_test_data/gfs.canopy.*_

drnimbusrain:
@zmoon  Thanks to Wei-Ting, I think the results of the new RSL_OPT=0 (Massman/MOST approach with lambdars) approach is working very well:   https://docs.google.com/presentation/d/1yRbnbYQQxkAnK79ud-n9MOw3V5YEcy55/edit#slide=id.p4

Therefore, can we move forward with review, and/or merge to develop?

angehung5:
Thanks Patrick @drnimbusrain , for posting the results. Should it be RSL_OPT=0?

Going to approve.

angehung5 (approved):

drnimbusrain:
@zmoon  What do you think about merge? I have just [cleaned up](https://github.com/noaa-oar-arl/canopy-app/pull/57/commits/721cb805e445f248657a5b78eb16ec3b90c5d404) the RSL_OPT=0 code to remove old fragmented RSL_OPT=1 codes.

README.md
```diff
@@ -183,9 +183,10 @@ https://nacc-in-the-cloud.s3.amazonaws.com/inputs/geo-files/canopy.2022MM01.test
 | `ifcanphot`     | logical canopy photolysis option (default: `.FALSE.`)                              |
 | `ifcanbio`      | logical canopy biogenic emissions option (default: `.FALSE.`)                      |
 | `href_opt`      | integer for using href_set in namelist (= `0`, default) or array from file (= `1`) |
-| `href_set`      | user-set real value of reference height above canopy associated with input wind speed (m) (only used if `href_opt=0`) |
+| `href_set`      | user-set real value of reference height above canopy associated with input wind speed (m) (only used if `href_opt=0`) ***\*\*** |
```
zmoon: It looks like only two are bolding, I think you want to use `**\*\*\***` instead

README.md
```diff
@@ -195,14 +196,15 @@ https://nacc-in-the-cloud.s3.amazonaws.com/inputs/geo-files/canopy.2022MM01.test
...
-| `rsl_opt`       | user-set option to include more explicit stability and Roughness SubLayer (RSL) effects in calculation of wind speed at canopy top (Uc) from reference height. `0`: off; `1`: on |
 | `bio_cce`       | user-set real value of MEGAN biogenic emissions "canopy environment coefficient" used to tune emissions to model inputs/calculations (default: `0.21`, based on Silva et al. 2020) |
 | `lai_thresh`    | user-set real value of LAI threshold for contiguous canopy (m2/m2)                 |
 | `frt_thresh`    | user-set real value of forest fraction threshold for contiguous canopy             |
 | `fch_thresh`    | user-set real value of canopy height  threshold for contiguous canopy (m)          |
 
 **\*\*** If `modres` >> `flameh` then some error in WAF calculation will be incurred.  Suggestion is to use relative fine `modres` (at least <= 0.5 m) compared to average flame heights (e.g., ~ 1.0 m) if WAF is required.
 
+***\*\*** If `href_set` becomes small and approaches z0 (or as `href_set` --> 0), only the sub-canopy wind profile is calculated, recommend `href_set` = 10 m.
```
zmoon: here too

zmoon (commented):
Noticed a little something in the readme

drnimbusrain:
Thank you Zach, can you fix it up?? ;)

drnimbusrain:
Thank you @zmoon !

zmoon (approved):
I didn't look closely but I say go for it

PR #58, opened 2023-04-13
Python examples

zmoon:
todo:
* [x] interactive var sel on the other 2-D case sets
* [x] refactor in some places a bit to reduce code duplication
  - made fn for interactive var sel
* [x] Binder config
* [x] run notebook(s) in CI
* [x] fix output var name/units consistency between text and nc outputs?
* [x] remove the Patrick notes from notebook


drnimbusrain:
@zmoon Thank you so much for doing all of this!  Understood that you should very soon turn your attention to UFS/NEXUS, but if we wrap this up it would be helpful!  Thanks again.

zmoon:
Looks like running the example nb currently takes ~ 50 s in CI. Not bad, but will later (probably not in this PR) add parallel sensitivity run option which will help with this.

zmoon:
Binder test link:

https://mybinder.org/v2/gh/noaa-oar-arl/canopy-app/py-examples?labpath=python%2Fexamples.ipynb

drnimbusrain:
@zmoon Thank you for continuing to work and wrap up these great Python examples and binder for the tests...I think it will be a great way for us to test canopy-app more effectively!

zmoon:
@drnimbusrain how do we want to address the var differences?

The bio emi ones are the same, the differences are:
| nc | txt |
|--- |--- |
| `canwind` | `ws` |
| `flameh` | `flame_height` |
| - | `canopy_height` |

Note that canopy height isn't in the nc output but is in the txt output.

The txt names are formed in the Python by taking those in the text output header and lowercasing and replacing ` ` with `_`.

drnimbusrain:
@zmoon This is a great point on the inconsistent variables.  I will add "canheight" to the NC output and change the txt output variable names to be consistent.

zmoon:
> I will add "canheight" to the NC output and change the txt output variable names to be consistent.

Ok, do you want to do it in this PR?

Or let me attempt and then you can review?

drnimbusrain:
@zmoon great to do it on this PR, and would appreciate if you could attempt then...

zmoon:
@drnimbusrain I'll remove your notes from the nb after you review, want to leave them for now so you have them.

zmoon:
I know I already changed it to `canwind`, but in retrospect maybe `ws` is better since WS above the canopy is included too, yes? Also fine with `canwind` though.

drnimbusrain:
Hi Zach, I am Ok with using WS for the canopy wind speed variable name.
Please proceed.

drnimbusrain (approved):

drnimbusrain:
@zmoon I think this looks great!  You can remove my notes and merge.

PR #77, opened 2023-06-29
add frp_src=3

angehung5:
FRP values are all ones when use ifcanwaf=.FALSE.


zmoon (commented):
Might as well change the comment in the Python readme too

angehung5:
Thanks Zach. Readme updated.

zmoon (approved):

PR #85, opened 2023-08-07
Update CI

zmoon:
* moving to the new micromamba Action
* [x] add more debug flags for GFortran
* [x] similar for Intel
* [x] DEBUG=2 option and document (29863dc9f)


zmoon:
@drnimbusrain how do we want to handle FPEs? First one is here (case that `hcmref` is 0):
```diff
--- a/src/canopy_calcs.F90
+++ b/src/canopy_calcs.F90
@@ -78,8 +78,12 @@ SUBROUTINE canopy_calcs
 ! ... calculate wind speed from u and v
                 ubzref   = sqrt((uref**2.0) + (vref**2.0))

-! ... get scaled canopy model profile and sub-canopy layers
-                zhc         = zk/hcmref
+                ! ... get scaled canopy model profile and sub-canopy layers
+                if (hcmref > 0) then
+                    zhc = zk / hcmref
+                else
+                    zhc = 0
+                end if
                 cansublays  = floor(hcmref/modres)

 ! ... check for valid model vegetation types
```

drnimbusrain:
I agree with your first FPE change here.  Do you want to make this (and other) change to your branch as the FPEs come up with your improved debugging options?

drnimbusrain:
@zmoon Do you think you will be able to work on this again soon?  Thank you!

zmoon:
@drnimbusrain I'll try to work on it again soon.

drnimbusrain:
Thanks Zach...I added the additional debug flags and checked/solved the
FPEs in latest develop.

zmoon:
That's good, I'll merge that into here before working on it more then.

zmoon:
@drnimbusrain some notes:
* CI debug build still has DEBUG=1 not 2
* we're currently not testing all code paths, just the default namelist, so there may be other FPEs
* "array temporary was created" warnings we could alleviate by explicitly creating the temporary arrays. This is mainly for going from an array of floats to assigning to a float member of an array of derived type in the ncf read/write.

drnimbusrain:
@zmoon Thank you for working on this, and agree that more comprehensive checks of the different code pathways/options are needed to check for all FPEs.  I will try and work on this with your new debug options more thoroughly as we go along with more robust testing.  I think that @btang1 @MaggieMarvin @angehung5 could help with this DEBUG=2 testing of the different options for FPEs as well, since they are going to be focusing on different parts of code.   For now I approve, and we can improve later.

drnimbusrain:
I can also work on the "array temporary was created" issues too.  Thanks!

src/canopy_utils_mod.F90
drnimbusrain, line 1: @zmoon Although, what has changed in the leafage option with your PR?

drnimbusrain (approved):
Looks OK, but just want to know about the leafage option changes here.

src/canopy_utils_mod.F90
zmoon: Should be nothing. I just moved the calculation under the if statement to avoid FPE related to the LAI being unitialized. Little note in d241beac46e

drnimbusrain:
OK, I will merge then and work with the DEBUG=2 option some to fix some other potential FPEs and the warnings...

PR #88, opened 2023-08-30
Added vegtype dependent FRP to FlameH calculations.

### LINKED ISSUE #86: [Fires] Add VegType Dependent FRP to Flame Height Calculation
Right now, we only use a single formation from Byram (1959) for calculation of FRP (intensity) to flame height relationship in [canopy-app](https://github.com/noaa-oar-arl/canopy-app/blob/develop/src/canopy_utils_mod.F90#L124).

We can/should use the work of[ Alexander and Cruz, Table 1 and S1 work.](https://www.publish.csiro.au/wf/pdf/WF11001_CO), and add vegetation type inputs to have a variable FRP to FlameH calculation. 

@angehung5

drnimbusrain:
This PR address Closes #86  , which stems from Wei-Ting's canopy-wind/WAF paper, and the work of [Alexander and Cruz (2012)](https://www.researchgate.net/publication/263022480_Interdependencies_between_flame_length_and_fireline_intensity_in_predicting_crown_fire_initiation_and_crown_scorch_height) (Table 1 and S1) that shows different vegetation type dependent relationships for deriving flame height based on fire intensity.

Previously we had used the single relationship based on Byram (1959) (also included in Table 1 of Alexander and Cruz (2012)) to derive flame height from FRP, and here we extend it to a methodology that is approximately related to model input vegetation type (currently supporting MODIS/VIIRS).

For the limited SE example using the FRP to flame height only option in canopy-app:

Byram (1959) Flame Heights
image

Alexander and Cruz (2012)  - Byram (1959) Flame Heights
image

@angehung5 When you return from leave, can you more robustly help test this new feature?


drnimbusrain:
After reading more of the Alexander and Cruz paper, I change this to a draft because there may be additional, more fuel-complex relationships to derive flame height, i.e., scorch height.

drnimbusrain:
I convert this to a draft, as when reading Alexander and Cruz, it may be more appropriate to use a vegtype dependent calculations of crown scorch height (hs), which can be related directly to flame height (hf) (see Table 2 and equation 14 in their paper).  I will add this option as well.

drnimbusrain:
I have added the additional option (```flameh_frp```) to use the vegtype dependent scorch height to flame height relationships from Alexander and Cruz.  

Overall, the differences compared to Byram (1959) only are similar as previous for the vegtype dependence, but the different relationship for grasslands/savanna/shublands/crops using scorch heights cause larger flame heights in this case for one grid cell:
image

drnimbusrain:
@angehung5 What do you think about this PR, will you give it a review?  Not sure it will have enough impact yet to warrant changing your paper results for midflame wind and WAF.  However, you mentioned you had comments on the new NL variable `flameh_frp`.  Also note that I have merged in the latest develop which contains multiple timesteps.  Thank you!

angehung5 (approved):
Sorry I haven't get a chance to test it, but I changed flameh_frp to flameh_cal (option for flameh calculation). Hope it is more clear.
You can go ahead and merge it to develop. I will find time to run some tests later.

PR #96, opened 2023-11-07
Feature/nl format

drnimbusrain:
@zmoon  @angehung5  Since we have over 40 NL options and growing, need to organize better.  I also added some python/README changes for running global data and an example slurm script to help.   Please let me know ASAP what you think so we can update develop going forward.


angehung5 (approved):

angehung5:
I changed the required cpu time and mem in the slurm script and readme based on my slurm settings on Hopper. Time might be a little off since I set ntask=12, but mem should be accurate (could be less depending on the case but it would be safer to put 60G).

python/README.md
```diff
@@ -96,17 +96,17 @@ You can also download and generate global gridded canopy-app inputs using Python
...
    #SBATCH --partition=bigmem             # big memory node
    #SBATCH --job-name=canopy-app          # name the job
    #SBATCH --output=canopy-%j.out         # write stdout to named file
    #SBATCH --error=canopy-%j.err          # write stdout to named file
-   #SBATCH --time=0-04:00:00              # Run for max of 00 hrs, 10 mins, 00 secs
+   #SBATCH --time=0-02:00:00              # Run for max of 00 hrs, 10 mins, 00 secs
    #SBATCH --nodes=1                      # Request N nodes
    #SBATCH --ntasks=1                     # Request n tasks
-   #SBATCH --mem-per-cpu=1000GB           # Request nGB RAM per core
+   #SBATCH --mem-per-cpu=60GB             # Request nGB RAM per core
```
drnimbusrain: @angehung5   A single node/task with your suggestion of 60GB memory may not work, please test this example.  

I don't think we really need to increase ntasks because its not parallel code.

From my experience using a single node/task requires like about 256 GB memory.

run_canopy_slurm.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -l
+#SBATCH --partition=bigmem             # big memory node
+#SBATCH --job-name=canopy-app          # name the job
+#SBATCH --output=canopy-%j.out         # write stdout to named file
+#SBATCH --error=canopy-%j.err          # write stdout to named file
+#SBATCH --time=0-02:00:00              # Run for max of 00 hrs, 10 mins, 00 secs
+#SBATCH --nodes=1                      # Request N nodes
+#SBATCH --exclude=hop006,hop010,hop011 # Exclude some nodes (optional)
+#SBATCH --ntasks=1                     # Request n tasks
+#SBATCH --mem-per-cpu=60GB             # Request nGB RAM per core
```
drnimbusrain: @angehung5 A single node/task with your suggestion of 60GB memory may not work, please test this example.

I don't think we really need to increase ntasks because its not parallel code.

From my experience using a single node/task requires like about 256 GB memory.

drnimbusrain (commented):
@angehung5 Please see my comments on your suggested memory updates here.

angehung5 (commented):
Tested on Hopper. For three time steps with node=1 and ntask=1:
```
Cores: 1
CPU Utilized: 01:12:51
CPU Efficiency: 95.37% of 01:16:23 core-walltime
Job Wall-clock time: 01:16:23
Memory Utilized: 9.57 GB
```
I put 12G just in case.

drnimbusrain:
> 1:16

@angehung5 OK, thank you.  Is this for running both the python global process script and canopy-app for those times?

angehung5:
> > 1:16
> 
> @angehung5 OK, thank you. Is this for running both the python global process script and canopy-app for those times?

global process script only.

drnimbusrain:
> > > 1:16
> > 
> > 
> > @angehung5 OK, thank you. Is this for running both the python global process script and canopy-app for those times?
> 
> global process script only.

@angehung5 Right, then the memory suggestion is not correct, and would need to be largely increased for canopy-app too.   We need to test run both at same time/job.

drnimbusrain:
Closing this PR due to updated new PR from @angehung5 built on latest [commit](https://github.com/noaa-oar-arl/canopy-app/pull/96/commits/d4801b947482967f76e6ae550aea5f5886210024).

PR #100, opened 2023-12-13
New WAI adaptation for PAI calculation option from Fang et al.

drnimbusrain:
@angehung5  This is the adaption of the WAI contribution using alpha (woody-to-total area ratio) for the PAI calculation option, PAI = LAI (1 - alpha), which is vegetation type dependent now.

This is based on Figure 1 of Fang et al., 2019:  https://doi.org/10.1029/2018RG000608

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ You can also [generate global inputs using Python (see python/global_data_proces
 | `z0ghc`         | ratio of ground roughness length to canopy top height (Massman et al., 2017)       |
 | `rsl_opt`       | user-set option for either MOST or unified Roughness SubLayer (RSL) effects above and at canopy top (Uc).(= `0`, default: uses MOST and a constant lambdars factor only), (= `1`, under development: will use a more unified RSL approach from Bonan et al. (2018) and Abolafia-Rosenzweig et al., 2021)   |
 | `lambdars`      | Value representing influence of RSL effects (with `rsl_opt=0`) (Massman et al., 2017)          |
-| `pai_opt`       | integer (`0`: PAI fixed from Katul et al. 2004 veg types-->default; `1`: PAI Massman et al. 2017 Eq. 19 calc; `2`: PAI from model LAI+WAI; `3`: user-set PAI value) |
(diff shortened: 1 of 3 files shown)


angehung5:
@drnimbusrain Are you using the Eq. 12 from Fang et al. 2019? If yes, how about the omega and gamma terms? Also should it be PAI = LAI / (1-alpha)? Otherwise, PAI would be less than LAI.

drnimbusrain:
> @drnimbusrain Are you using the Eq. 12 from Fang et al. 2019? If yes, how about the omega and gamma terms? Also should it be PAI = LAI / (1-alpha)? Otherwise, PAI would be less than LAI.

Actually, I that is what I am using PAI = LAI/(1-alpha):  https://github.com/noaa-oar-arl/canopy-app/pull/100/files#diff-784b83d59c46a5575c78286c41ec9665e53c5edad0c9a4bc4e97e4abdb6cb743R62

angehung5:
> > @drnimbusrain Are you using the Eq. 12 from Fang et al. 2019? If yes, how about the omega and gamma terms? Also should it be PAI = LAI / (1-alpha)? Otherwise, PAI would be less than LAI.
> 
> Actually, I that is what I am using PAI = LAI/(1-alpha): https://github.com/noaa-oar-arl/canopy-app/pull/100/files#diff-784b83d59c46a5575c78286c41ec9665e53c5edad0c9a4bc4e97e4abdb6cb743R62

How about the clumping index (omega term)? The paper says "ΩE is the element clumping index, which quantifies the effect of foliage clumping at scales larger than shoots". Should it be included in the calculation? 
Yeah I think we can ignore gamma (needle‐to‐shoot area ratio) since we are working on grown trees.

drnimbusrain:
@angehung5  Ah, yes, there are some other details here with gamma_eff (needle-to-shoot ratio, i.e., effect of clumping within the shoot) and omega_eff (clumping index effect beyond the shoot).  
 However we have observed total omega, clumping index (within and beyond the shoot) which is defined by  (see Eq. 16 here https://academic.oup.com/treephys/article/29/8/1069/1670149_

Omega = Omega_eff/gamma_eff    ;  Omega)_eff is the clumping beyond the shoot only, where the gamma_eff is the effect of clumping within the shoot.  

So combining this with Zhang et al.  PAI_eff = LAI/(1-alpha) * Omega_eff/gamma_eff, we have a final equation for PAI total (not effective)

**PAI = LAI/(1-alpha) * Omega.**
So I do need to revise this for explanation and to include the effects of total clumping index here to get PAI.

drnimbusrain:
@angehung5 OK, updated.

angehung5 (approved):

angehung5:
@drnimbusrain Thanks Patrick! Just reviewed.

PR #101, opened 2024-01-12
Separated SSG from crops in foliage distribution and other vtypes.

drnimbusrain:
Separating shrublands somewhat and basing its distribution more on forests following Bill's comments.  As you noted and like in Bonan, the grasslands/savannas are still treated similar to croplands.  I also noticed I should add additional possible vtypes from MODIS/VIIRS, including Cat=14 croplands/mosaic (treated same as croplands), and Cats18-19 wooded/mixed tundra (treated same as low-lying shrublands).


angehung5 (approved):

drnimbusrain:
Thank you @angehung5 for reviewing, I will merge now.  In your updated develop branch testing can you please check and make sure it all relatively makes sense and nothing is obviously broken?

PR #102, opened 2024-01-17
More explicit M17 way of calculating Uc with Eq. 10 and 14.

drnimbusrain:
This method simply combines Eq. 10 and 14 from Massman et al. 2017 to calculate the Uc, which may result in changes too.  @angehung5 Can you test this?

### DIFF
--- a/src/canopy_calcs.F90
+++ b/src/canopy_calcs.F90
@@ -190,7 +190,7 @@ SUBROUTINE canopy_calcs(nn)
                                     do k=1, modlays
                                         call canopy_wind_most(hcmref, zk(k), fafraczInt(k), ubzref, &
                                             z0ghc, cdrag, pai, hgtref, d_h, zo_h, &
-                                            lambdars, canBOT(k), canTOP(k), canWIND_3d(i,j,k))
+                                            lambdars, molref, canBOT(k), canTOP(k), canWIND_3d(i,j,k))
                                     end do
                                 else
                                     write(*,*) 'wrong RSL_OPT namelist option = ', rsl_opt, 'only option = 0 right now'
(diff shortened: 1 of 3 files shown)


angehung5:
Below is the comparison of wind profiles using original calculation and M17 formulation. M17 does better at Uc estimation and probably better at in-canopy wind, but the above-canopy wind speeds are way higher. Also the discontinuity at the canopy top is more significant when using M17. 

I am not sure which approach is better. The original calculation has larger bias in terms of Uc but the overall profile is ok and more smooth. M17 has better performance in Uc but the shape of the overall wind profile is kind of strange. 

xTE_wind_profile_0600_20200601_20200930_avg_sensitivity_uc_zh

drnimbusrain:
Seems new M17 approach is overall better here,  and I don't see a greater
discontinuity (maybe a bit sharper).  What do other sites show, and for
those that have above canopy?  I agree its definitely higher in new M17
approach.

Thanks

angehung5:
Seems the Uc estimation based on M17 works fine at Bartlett forest but is much higher at Chestnut, GoAmazon and Willow Creek, as well as above-canopy wind speeds. I would suggest sticking to current approach. 

Presentation1

drnimbusrain:
@angehung5 OK, I agree, its too high.  Sorry, but made another slight change to this M17 approach, which should lower it now.  Can you check once more these results at four sites?  Thank you!

drnimbusrain:
Sorry, made another change for the zpd calculation to be consistent with updated M17 method.  Please check now after latest commit.

drnimbusrain:
@angehung5  Closing this for now, as we will remain with original approach for now, and work on this in future for more robust unified stability/RSL approach following Bonan.

PR #103, opened 2024-01-19
Change YY to 2022 for gfs.canopy in global script

angehung5:


drnimbusrain (approved):
Looks good, seems you just needed to hard code the dates of the daily canopy files.

PR #104, opened 2024-01-24
Hoping to Merge Develop Changes to Stable

drnimbusrain:
@zmoon


README.md
```diff
@@ -149,10 +150,12 @@ The Canopy-App input data in [Table 2](#table-2-canopy-app-required-input-variab
 | `hpbl`                           | Height of the planetary boundary layer (m)  | UFS NOAA/GFSv16                                    |
 | `prate_ave`                      | Average mass precipitation rate (kg m-2 s-1) | UFS NOAA/GFSv16                                   |
 | **External Canopy Variables**    | **Variable Description and Units**          | **Data Source/Reference (if necessary)**           |
-| `fh`                             | Forest canopy height (m)                    | Fused GEDI/Landsat data. Data Period=2020. ([Potapov et al., 2020](https://doi.org/10.1016/j.rse.2020.112165)) |
-| `clu`                            | Forest clumping index (dimensionless)       | GriddingMachine/MODIS. Data Period=2001-2017 Climatology. ([Wei et al., 2019](https://doi.org/10.1016/j.rse.2019.111296)). Extended globally for high latitudes using methods described [here](https://gmuedu-my.sharepoint.com/:w:/g/personal/whung_gmu_edu/EdglXmW2kzBDtDj1xV0alGcB1Yo2I8hzdyWGVGB2YOTfgw). |
```
angehung5, line 153: Maybe use "Vegetation clumping index"? It should be clu for all species.

angehung5 (approved):
I did not check the bioemi and ipynb scripts. Others look good. But I suggest changing "Forest clumping index" to "Vegetation clumping index" in README, since it should include all vegetation species.

drnimbusrain:
Thank you Wei-Ting! This brings up the question again on if we should make
the change from MODIS forest fraction to total vegetation fraction.

drnimbusrain:
@angehung5  Did some checking, currently the ffrac is only used to define [contiguous canopy grid cell threshold](https://github.com/noaa-oar-arl/canopy-app/blob/develop/src/canopy_calcs.F90#L139), and in the [PAI calculation option as a surrogate for canopy cover fraction](https://github.com/noaa-oar-arl/canopy-app/blob/develop/src/canopy_utils_mod.F90#L71) (but this M17 formulation/option leads to results we do not trust).   

Overall, since canopy-app is technically being driven by any canopy type (for height, lai, clu, etc.) it seems to me that we should change this "_ffrac_" variable into our canopy-app inputs to rather use a total canopy fraction, i.e.,  "_canfrac_".  It would be more general, and likely more representative as a threshold (and in the PAI M17 calculation).  We could add this canfrac also as an additional 1-km global file for UFS later too.

I don't think this effects much your paper results, but could be something that could be brought in here.  As you mentioned, this is something you already have as the annual total vegetation fraction (TVFRAC) in your [global extension document](https://gmuedu-my.sharepoint.com/:w:/g/personal/whung_gmu_edu/EdglXmW2kzBDtDj1xV0alGcB1Yo2I8hzdyWGVGB2YOTfgw), and something that was needed for ffrac anyway.

What do you think?

README.md
drnimbusrain: Agreed and changed to canopy clumping index.

drnimbusrain:
@angehung5  See issue #106   I think we should also make this change for 'ffrac' to 'canfrac' before merging develop to stable...

PR #107, opened 2024-02-02
Fixed array temporary warnings by explicitly creating the arrays.

drnimbusrain:
@zmoon could you have a quick look, I have explicitly set all arrays in netcdf read and when they are also used in some subroutine calls in canopy_calcs...This effectively removes all array temporary warnings when DEBUG=2.


zmoon (approved):
Looks fine to me

PR #117, opened 2024-02-13
Adding LAD as output

### LINKED ISSUE #116: [Outputs]  Add Calculated Leaf Area Density as Output

drnimbusrain:


drnimbusrain:
@angehung5 Can you have a quick review of this?  I added a calculation of LAD as output (from shape distribution function) to .txt and netcdf.  Thought it could be useful to have the approximate LAD in the output.  Also, this may be insightful information for @MaggieMarvin biogenic emissions comparisons against sites.

src/canopy_init.F90
```diff
@@ -40,6 +42,8 @@ SUBROUTINE canopy_init
         if(allocated(canWIND_3d))    canWIND_3d(:,:,:)    = fillreal
         if(allocated(dx))            dx(:)                = fillreal
         if(allocated(dx_2d))         dx_2d(:,:)           = fillreal
+        if(allocated(flameh))        flameh(:)            = fillreal
```
drnimbusrain, line 45: I also added initializations for flameh as they are needed in the output...

src/canopy_init.F90
```diff
@@ -40,6 +42,8 @@ SUBROUTINE canopy_init
...
+        if(allocated(flameh_2d))     flameh_2d(:,:)       = fillreal
```
drnimbusrain, line 46: I also added initializations for flameh as they are needed in the output...

src/canopy_ncf_io_mod.F90
```diff
@@ -804,7 +819,9 @@ SUBROUTINE canopy_outncf_alloc
         ! Time-varying 2d fields at cell centers.
         !-------------------------------------------------------------------------------
 
-        nfld2dxyt = 1  ! canopy height
+        nfld2dxyt = 0
+
+        nfld2dxyt = nfld2dxyt +1  ! canopy height
```
drnimbusrain, line 824: Slight change here to be clear on indices for each variable added to output...

angehung5 (approved):

PR #120, opened 2024-02-16
Canopy loss factor.

### LINKED ISSUE #119: [Biogenics] Adding Canopy Loss Factor
@MaggieMarvin  I will add an approximate canopy loss factor when bio summing option is used.  

It is based on Eq. 21 of [Guenther et al. 2006](https://acp.copernicus.org/articles/6/3181/2006/))

This will facilitate improved comparison of canopy-app top primary biogenic emissions from top of canopy with ground-truth flux observations (e.g., PECORINO, FLUCs, etc.) that inherently measure fluxes after chemistry and deposition loss have occurred in-canopy.  We can always turn this off when directly using leaf-level canopy-app emissions with chemical box models that may have BVOC chemistry and deposition (e.g., F0AM)

drnimbusrain:
@MaggieMarvin Here is an approximate canopy loss factor when bio summing option is used.

It is based on Eq. 21 of [Guenther et al. 2006](https://acp.copernicus.org/articles/6/3181/2006/))

This will facilitate improved comparison of canopy-app top primary biogenic emissions from top of canopy with ground-truth flux observations (e.g., PECORINO, FLUCs, etc.) that inherently measure fluxes after chemistry and deposition loss have occurred in-canopy. We can always turn this off when directly using leaf-level canopy-app emissions with chemical box models that may have BVOC chemistry and deposition (e.g., F0AM)


MaggieMarvin (approved):
These changes work as expected - option 1 varies with canopy height and reduces emissions with loss factors on the order of 0.99 (see attached plot), while option 2 produces a constant reduction of 0.96 at all emission locations. It's good that we've added more warnings and options until we can gain some more clarity on the application of lambda to different PFTs and to species other than isoprene.
map_canopyapp_bio_NA-28

drnimbusrain:
@MaggieMarvin Great, thanks.  I will wait to merge though, as I'd rather get historical changes in first and then resolve the conflicts only once here.

MaggieMarvin (approved):

PR #128, opened 2024-03-21
Add biogenic species selection option for NetCDF output files

MaggieMarvin:
To help reduce the size of global NetCDF output files (with 25+ hour spinup, global gridded output files can be on the order of 1 TB), I've added the option to choose whether to output biogenic emissions for all species or just one as identified by its corresponding ID number (Table 1). Tables 1 and 3 are updated in the README. Does not currently support multiple IDs or apply to txt output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,27 +89,27 @@ Current Canopy-App components:
 
 ### Table 1. Canopy-App Biogenic Emissions Output Variables
 
-| Variable Name | Variable Description (Units: kg m-3 s-1)  |
-| ------------- | ----------------------------------------- |
-| `emi_isop`    | Isoprene                                  |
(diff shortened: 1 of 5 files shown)


drnimbusrain (approved):
@MaggieMarvin I think this is a nice addition here!  I am in favor of approving, so as to help facilitate your further testing of biogenics.  I do think maybe at some point later we could add a multi-species/multi-index  output option entering an array of indices or strings so that the model wouldn't need to be re-run for different species each time.  It would be relatively easy to change biospec_opt to be an array of integers, and add a loop through biospec_opt in canopy_ncf_io_mod.F90 to use all indices input.

MaggieMarvin:
Agreed, I think a multi-species output option would be useful too!

PR #129, opened 2024-03-22
Changing to individual time step outputs for NetCDF files.

drnimbusrain:
@MaggieMarvin To avoid outputting to an overly large NetCDF output file in canopy-app, I decided to just change it to similarly how the text files work.  Now, here by default, canopy-app will output separate NetCDF files for each timestep (e.g., hourly) with the output timestep stamp on the filenames.   Its not the prettiest right now, but this should further facilitate testing longer biogenic simulations instead of having a single huge file.  Please have a close look and make sure the outputs look good.


MaggieMarvin (approved):
Confirmed that this works as expected - for the standard test case that is provided with canopy-app, this generates three output files, each containing the information expected for that time step t00*. This will definitely help with file size and is great for now! I'm not sure how easy it would be but at some point later I think it would help to change the output file naming convention from t00* to the corresponding time stamp if possible.

drnimbusrain:
> Confirmed that this works as expected - for the standard test case that is provided with canopy-app, this generates three output files, each containing the information expected for that time step t00*. This will definitely help with file size and is great for now! I'm not sure how easy it would be but at some point later I think it would help to change the output file naming convention from t00* to the corresponding time stamp if possible.

@MaggieMarvin I agree with you.  I have now changed both NCF and TXT outputs to instead append the valid time stamp to the output files now.  Can you please check?

MaggieMarvin:
Yes that works as expected, thank you!

Issue #124, opened 2024-03-13
[biogenics]  Add air quality/ozone stress gamma based on GFSv16


drnimbusrain:
@MaggieMarvin   This should be relatively simple addition, where we can add the MEGAN3 function for calculating "GAMMA_AQ", which depends on a current air quality index, i.e. in MEGAN3 it uses ozone W126 (ppm hours-1).  This can be calculated based on input ozone concentrations.  Initially, we can make an option for a a set constant ozone mixing ratio input in the namelist and calculate the [W126](https://www.epa.gov/sites/default/files/2015-09/documents/w126_steps_to_calculate_revised_feb19.pdf) (easier) in canopy-app, but it should be spatially dependent for regional and global scale applications, thus calculated using gridded input ozone concentrations to canopy-app (trickier).

drnimbusrain:
Also, realized that we can add the GFS ozone mixing ratio from the 3D files, gfs.t12z.atmf*.nc (variable name = 'o3mr'), which are available on AWS as similar to the surface files (i.e.,gfs.t12z.sfcf*.nc) .  Maybe @angehung5 can help us add these first model layer ozone concentrations to the example and global canopy inputs, and update her python [global_data_process.py](https://github.com/noaa-oar-arl/canopy-app/blob/develop/python/global_data_process.py) script, we should be able to have this spatially dependent calculation of W126 ozone for input into the air quality/ozone stress gamma index.  I think we should incorporate this to canopy-app.

drnimbusrain:
The main issue, however, is that the W126 ozone index is based on a long-term averaging, not instantaneous GFS ozone mixing ratios.  Therefore, may be best to extract the 1st model layer ozone from GFS for 2021 - 2024 and calculate the  [W126](https://www.epa.gov/sites/default/files/2015-09/documents/w126_steps_to_calculate_revised_feb19.pdf) as an offline pre-process step and add the effective climatological index to a new "canopy" variable to the example files and AWS.

drnimbusrain:
[...]

PR #137, opened 2024-07-08
Add ozone W126

angehung5:
#124 
SE example files updated. README is also updated by adding a brief description of w126. Please check if that is correct.


drnimbusrain:
@angehung5   do you think it's ok to leave values over ocean as zeros? Or should I replace with nans?  How does the other canopy variable treat this?  This is a question from @MaggieMarvin

README.md
```diff
@@ -170,6 +170,7 @@ The Canopy-App input data in [Table 2](#table-2-canopy-app-required-input-variab
 | `csz`                            | Cosine of the solar zenith angle (dimensionless) | [Based on Python Pysolar](https://pysolar.readthedocs.io/en/latest/) |
 | `mol`                            | Monin-Obukhov Length (m)                    | Externally calculated using GFS `tmp2m`, `fricv`, and `shtfl`.  ([Essa, 1999](https://inis.iaea.org/collection/NCLCollectionStore/_Public/37/118/37118528.pdf)) |
 | `href`                           | Reference height above canopy (m) - 10 m    | Assumed constant (i.e., 10 m).  Can be taken from NL. |
+| `ozone_w126`                     | Ozone W126 index (ppm-hours)                | Climatological value during 04/2021-04/2024 based on GFSv16 lowest model later ozone mixing ratios. |
```
drnimbusrain:
```suggestion
| `ozone_w126`                     | Ozone W126 index (ppm-hours)                | Climatological value during 04/2021-04/2024 based on GFSv16 lowest model layer ozone mixing ratios. |
```

drnimbusrain:
@angehung5 Also, even though I haven't added the canopy-app code changes to read in W126, does your changes here work ok in testing both default NC and TXT examples with the new data there?

drnimbusrain:
@MaggieMarvin I think you should also review these changes quickly.

angehung5:
@MaggieMarvin It should be ok to leave oceans as zeros. I will change them to zeros anyway since nans may break the code, unless you want to put other values (e.g. -999). 

@drnimbusrain Yes, it should work. I just add a new variable to the files, nothing else changes.

input/point_file_20220630.sfcf023.txt
```diff
@@ -1,2 +1,2 @@
-lat,lon,ch,ugrd10m,vgrd10m,clu,lai,vtype,canfrac,fricv,csz,sfcr,mol,frp,href,sotyp,pressfc,dswrf,shtfl,tmpsfc,tmp2m,spfh2m,hpbl,prate_ave,soilw1,soilw2,soilw3,soilw4,wilt
-34.03,272.11,20.8692,-0.7485,0.3286,0.5214,3.7059,4,0.2966,0.1107,0.0416,0.8260,12.2366,0.0000,10.00,3,99731.6094,10.1607,-9.8159,292.9766,293.2202,0.0146,14.0288,0.0001,0.2591,0.2053,0.2134,0.2132,0.0470
+lat,lon,ch,ugrd10m,vgrd10m,clu,lai,vtype,canfrac,fricv,csz,sfcr,mol,frp,href,sotyp,pressfc,dswrf,shtfl,tmpsfc,tmp2m,spfh2m,hpbl,prate_ave,soilw1,soilw2,soilw3,soilw4,wilt,ozone_w126
+34.03,272.11,20.8692,-0.7485,0.3286,0.5214,3.7059,4,0.2966,0.1107,0.0416,0.8260,12.2366,0.0000,10.00,3,99731.6094,10.1607,-9.8159,292.9766,293.2202,0.0146,14.0288,0.0001,0.2591,0.2053,0.2134,0.2132,0.0470,85
```
zmoon, line 2: Is it always rounded to the closest integer?

drnimbusrain:
Good point Zach,  can you check Maggie and Wei-Ting

angehung5:
According to EPA's definition (https://www.epa.gov/sites/default/files/2015-09/documents/w126_steps_to_calculate_revised_feb19.pdf), W126 could be floating numbers but they all integers in the file `/groups/NA22OAR/pcampbe8/gfsv16_ozone_w126/gfsv16_ozone_w126_042021-042024.nc`.
@MaggieMarvin Could you double check? Thanks!

MaggieMarvin (approved):
Looks great to me! The input NC files have correct values and the test case runs just fine - just needs typo fix "later" to layer in README (line 173)

MaggieMarvin:
Thanks Wei-Ting! In the very last step the guidance is to round to the nearest whole number so that's what I've done here

drnimbusrain:
Great Maggie!  Does TXT test case also still work fine?

MaggieMarvin:
Yes! For both gridded and point test cases

drnimbusrain (approved):
just small typo correction.  approved.

drnimbusrain:
@angehung5 @MaggieMarvin I merged, but seeing some strange values in the SE example files:

image

PR #138, opened 2024-07-09
Update W126 using new timezone map

angehung5:
#124 
Weird pattern due to land/water mask fixed by updating the timezone map. It looks better now!

Screenshot 2024-07-09 at 1 00 25 PM


drnimbusrain:
@angehung5 Great news!  How come only three files are updated here?

angehung5:
For some reasons, it does not let me replace certain files. I removed all example files and reuploaded. It should work now.

drnimbusrain (approved):

PR #142, opened 2025-01-28
Feature/calc temp press hum

drnimbusrain:
@btang1   In preparation for adding initial dry deposition scheme (Zhang et al., 2002) from ACCESS, I add simply here first approximations (no leaf energy balance...yet) from ACCESS for ambient temperature (tka), pressure (pressa), relative humidity (relhuma), and specific humidity (spechuma) within and above canopy.  This must use an approximation of initially constant specific humidity (based on 2-meter AGL specific humidity from model/obs) to calculate vertically varying relhuma, which then is used to calculate spechuma with height.  

The single point test results are simple linearly decreasing functions for tka and pressa with height, linearly increasing relhuma with height (with decreasing tka under assumed constant spechuma), and decreasing spechuma with height above point where to satisfy 99% relhuma conditions. 
image
image
image
image

The variables are needed in canopy-app for calculation of Zhang et al. drydep method, which will be added later after confirmation of this approach.


drnimbusrain:
@btang1 @zmoon no worries if you have no time to review.  This is a    very simple straightforward approx to get this started.  I may merge.

zmoon (commented):
How about changing the RH and spec hum func names to start with "Calc" as well?

drnimbusrain:
> How about changing the RH and spec hum func names to start with "Calc" as well?

Thanks @zmoon !

drnimbusrain:
@zmoon @btang1 If there are no objections, I will merge by COB today.  Thanks again!

PR #144, opened 2025-02-10
add features to output d_h and zo_h

angehung5:


src/canopy_ncf_io_mod.F90
```diff
@@ -369,6 +369,30 @@ SUBROUTINE canopy_outncf_init
         c_canheight%iend(1) = nlon
         c_canheight%iend(2) = nlat
 
+        c_dh%fld = fillreal
+        c_dh%fldname = 'dh'
+        c_dh%long_name = 'zero-plane displacement height (z/h)'
+        c_dh%units = 'm'
```
zmoon: if it's relative to canopy height it shouldn't be m here

angehung5: You are right. Thanks for reminding!

zmoon:
I'll have to investigate more why the notebook is not working with this.

Also, I notice this warning:
```
canopy_calcs.F90:1010:61:

 1010 |                             z0ref, vtyperef, lu_opt, z0_opt, dh_2d(i,j), zoh_2d(i,j))
      |                                                             ^
Warning: ‘j’ may be used uninitialized [-Wmaybe-uninitialized]
canopy_calcs.F90:1010:61: Warning: ‘i’ may be used uninitialized [-Wmaybe-uninitialized]
```
(calling `canopy_zpd`)

Edit: I think it's just related to this, the first 1-D run example fails at runtime.

src/canopy_calcs.F90
```diff
@@ -1007,7 +1007,7 @@ SUBROUTINE canopy_calcs(nn)
 
                         call canopy_zpd(zhc(1:cansublays), fafraczInt(1:cansublays), &
                             ubzref, z0ghc, lambdars, cdrag, pai, hcmref, hgtref, &
-                            z0ref, vtyperef, lu_opt, z0_opt, d_h, zo_h)
+                            z0ref, vtyperef, lu_opt, z0_opt, dh_2d(i,j), zoh_2d(i,j))
```
zmoon: This is in the `loc` loop, not the `i`,`j` loop. Probably want to use the 1-d versions you made?

drnimbusrain: @angehung5 I agree with Zach, here and below in the 1D part of these codes it should be d_h(loc) and zoh(loc).  
```suggestion
                            z0ref, vtyperef, lu_opt, z0_opt, d_h(loc), zoh(loc))
```
This likely has to do with the compilation warnings too since i,j are not used, but rather loc in the 1D routines.  Please ensure that you are compiling/running with full DEBUG=2 on as well to best check everything. 

```
Warning: ‘j’ may be used uninitialized [-Wmaybe-uninitialized]
canopy_calcs.F90:1010:61: Warning: ‘i’ may be used uninitialized [-Wmaybe-uninitialized]
```

Also, please and test/confirm the 1D versions of these outputs function normally.  Since we don't have too many variables yet, maybe we need to add these d_h and zoh as default into the 1D text output for all options ([maybe similarly to as LAD or the 2D WAF](https://github.com/noaa-oar-arl/canopy-app/blob/develop/src/canopy_txt_io_mod.F90#L113))?

src/canopy_calcs.F90
```diff
@@ -1029,7 +1029,7 @@ SUBROUTINE canopy_calcs(nn)
                             if (rsl_opt .eq. 0) then
                                 do k=1, modlays
                                     call canopy_wind_most(hcmref, zk(k), fafraczInt(k), ubzref, &
-                                        z0ghc, cdrag, pai, hgtref, d_h, zo_h, &
+                                        z0ghc, cdrag, pai, hgtref, dh_2d(i,j), zoh_2d(i,j), &
```
drnimbusrain:
```suggestion
                                        z0ghc, cdrag, pai, hgtref, d_h(loc), zoh(loc), &
```

src/canopy_calcs.F90
```diff
@@ -1045,7 +1045,7 @@ SUBROUTINE canopy_calcs(nn)
                                 if (flameh(loc) .gt. 0.0) then !flameh must be > 0
                                     if (flameh(loc) .le. hcmref) then !only calculate when flameh <= FCH
                                         call canopy_waf(hcmref, lambdars, hgtref, flameh(loc), &
-                                            firetype, d_h, zo_h, canBOT(midflamepoint), &
+                                            firetype, dh_2d(i,j), zoh_2d(i,j), canBOT(midflamepoint), &
```
drnimbusrain:
```suggestion
                                            firetype, d_h(loc), zoh(loc), canBOT(midflamepoint), &
```

src/canopy_calcs.F90
```diff
@@ -1055,7 +1055,7 @@ SUBROUTINE canopy_calcs(nn)
                             else  !grass/crops, above-canopy firetype
                                 if (flameh(loc) .gt. 0.0) then !flameh still must be > 0
                                     call canopy_waf(hcmref, lambdars, hgtref, flameh(loc), &
-                                        firetype, d_h, zo_h, canBOT(midflamepoint), &
+                                        firetype, dh_2d(i,j), zoh_2d(i,j), canBOT(midflamepoint), &
```
drnimbusrain:
```suggestion
                                        firetype, d_h(loc), zoh(loc), canBOT(midflamepoint), &
```

src/canopy_ncf_io_mod.F90
```diff
@@ -369,6 +369,30 @@ SUBROUTINE canopy_outncf_init
         c_canheight%iend(1) = nlon
         c_canheight%iend(2) = nlat
 
+        c_dh%fld = fillreal
+        c_dh%fldname = 'dh'
+        c_dh%long_name = 'zero-plane displacement height (z/h)'
```
drnimbusrain: @angehung5 This is actually the ratio of zero plane displacement height to canopy height, d/h.
```suggestion
        c_dh%long_name = 'ratio of zero-plane displacement height to canopy height (d/h)'
```

src/canopy_ncf_io_mod.F90
```diff
@@ -369,6 +369,30 @@ SUBROUTINE canopy_outncf_init
...
+        c_dh%dimnames(1) = 'nlon'
+        c_dh%dimnames(2) = 'nlat'
+        c_dh%istart(1) = 1
+        c_dh%istart(2) = 1
+        c_dh%iend(1) = nlon
+        c_dh%iend(2) = nlat
+
+        c_zoh%fld = fillreal
+        c_zoh%fldname = 'z0h'
+        c_zoh%long_name = 'surface roughness length (z/h)'
```
drnimbusrain:
```suggestion
        c_zoh%long_name = 'ratio of surface roughness length to canopy height (z/h)'
```

drnimbusrain (changes requested):

src/canopy_alloc.F90
```diff
@@ -45,6 +45,10 @@ SUBROUTINE canopy_alloc
     if(.not.allocated(ppfd_ave))           allocate(ppfd_ave(modlays))
     if(.not.allocated(lad))                allocate(lad(nlat*nlon,modlays))
     if(.not.allocated(lad_3d))             allocate(lad_3d(nlon,nlat,modlays))
+    if(.not.allocated(zo_h))               allocate(zo_h(nlat*nlon))
+    if(.not.allocated(zoh_2d))             allocate(zoh_2d(nlon,nlat))
+    if(.not.allocated(d_h))                allocate(d_h(nlat*nlon))
+    if(.not.allocated(dh_2d))              allocate(dh_2d(nlon,nlat))
```
zmoon: It's a bit confusing that the names don't match between the 1-d and 2-d versions, with the 1-d using `_` before h but the 2-d not. Maybe pick one way to do it.

drnimbusrain:
Yes, agreed.  Let's make them consistent

angehung5:
Sorry I forgot to check the 1D routines, only test the 2D nc files. It should work now.
Also add d_h and z0_h to 1D txt.

angehung5:
Try running canopy-app with global gfs with the updated code and get this: `In file 'canopy_alloc.F90', around line 68: Error allocating 909744537600 bytes: Cannot allocate memory`

Diffusivity, photolysis, and biogenic emission options are off. Any other suggestions?

zmoon:
> 909744537600 bytes

wow almost 1 TB. Can you change the `hist_opt` so those things are not allocated? How many points are you running?

angehung5:
I am running 1536x3072 points. Model works well after setting `hist_opt` to 0.

drnimbusrain:
> canopy_alloc.F90

Thank you for checking this Wei-Ting, and clearly our biogenic historic option needs help as it requires a lot of memory.  Seems I need an  `if (ifcanbio)` around those historical blocks of tmp variables too as they do not need to be allocated unless that option is on.

drnimbusrain:
Or maybe not since technically these calculations are done before biogenics and could be used elsewhere...  @MaggieMarvin What do you think?  Seems like we need to disable hist_opt at least as default off (=0) in example namelist.  I think we could move the historical calculations into the ifcanbio block of canopy_calcs and also move those alloc/dealloc variables inside the ifcanbio block as well though.  We can address in future PR with your biogenic updates coming.

drnimbusrain:
> I am running 1536x3072 points. Model works well after setting `hist_opt` to 0.

@angehung5 So you have global outputs working ok for z0/h and d/h?

angehung5:
@drnimbusrain Yes, they seem working fine. Haven't compare to gfs and/or satellite retrievals but the values look reasonable.

Screenshot 2025-02-13 at 3 32 04 PM
Screenshot 2025-02-13 at 3 33 25 PM

drnimbusrain:
> @drnimbusrain Yes, they seem working fine. Haven't compare to gfs and/or satellite retrievals but the values look reasonable.
> 
> Screenshot 2025-02-13 at 3 32 04 PM Screenshot 2025-02-13 at 3 33 25 PM

Wow, cool, nice to see its working.  I will review.

drnimbusrain (approved):

PR #145, opened 2025-02-11
Update CI

zmoon:
The `python canopy_app.py` step should've failed in #144 initially but didn't. This should fix that and also turns on verbosity.

Also updated versions of things used in the Workflows.


drnimbusrain (approved):

drnimbusrain:
@zmoon Thanks!  Go ahead and merge.

PR #150, opened 2025-02-27
Add LDF/LIF to emissions calculations

MaggieMarvin:
These changes account for the light-dependent fraction (LDF) and light-independent fraction (LIF) in our leaf-level calculations of biogenic emissions. For each species, the LDF describes the fraction of emissions that requires sunlight to form and the LIF is instead totally dependent on temperature. Although these changes are not expected to impact isoprene which is fully light dependent (LDF = 1), most other species have a lower LDF and a greater fraction of emissions that are more sensitive to temperature.

I've implemented the LDF/LIF following MEGANv3 which is slightly different from Guenther et al. (2012) and involves combining the gammas for temperature and PAR together first before inserting into the final calculation. This is to avoid multiplying by a gamma PAR of zero and otherwise adjusting gamma PAR arbitrarily. For the LIF part of the calculation I also use a reference temperature value of 303 K, which is used also in MEGANv3 as well as MEGANv2.

I also found that in calculating LIF, I matched observations best when using air temperature versus the leaf temperature recommended by Guenther et al. (2012) - otherwise emissions were much too high. MEGANv2 also uses air temperature (TEMP2M) for this calculation, but MEGANv3 uses leaf temperature as estimated via energy balance. Since emissions between MEGANv2 and MEGANv3 seem very similar for the cases I looked at, this suggests that our current leaf temperature estimation in canopy-app might be too high. Something to revisit in the future but for now I've split the difference by using air temperature interpolated to each canopy layer.

Results for monoterpenes from the PECORINO campaign are shown below: now MT emissions continue overnight even when sunlight is not present and match observations as well as results from MEGANv2 and MEGANv3 very well.

point_timeseries_plot_long-107


drnimbusrain:
@MaggieMarvin Great work!  However, please make sure to always develop/test in DEBUG mode with new changes. Before I review, I just compiled and ran your branch for default NL case in debug mode (i.e., DEBUG=2 NC=1 make) and there are floating point exception that is associated with biogenic codes (not the case in upstream develop branch):

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
...
**/src/canopy_bioemi_mod.F90:226**

Likely this is the case with not having the daytime only limits on calculating biogenic emissions (https://github.com/noaa-oar-arl/canopy-app/pull/150/files#diff-cf677e4181ec06af78afafeacfc0ab0e60b76e38ba28d17a5f1b45c499bc814cL485-L486), which is good now especially considering new LDF/LIF impacts and emissions at night, but we will need to have conditions inside of the biogenic emissions module to deal with this.  This is likely the case only in short runs (< 24 hours) using instantaneous leaf PPFD during nightime hours in certain regions of domain, but need to fix regardless.

src/canopy_calcs.F90
```diff
@@ -1273,15 +1272,14 @@ SUBROUTINE canopy_calcs(nn)
                         end if
 ! ... user option to calculate in-canopy biogenic emissions
                         if (ifcanbio) then
-                            if (cszref .ge. 0.0_rk .and. dswrfref .gt. 0.0_rk &
```
drnimbusrain, line 1276: Compiling/running in DEBUG mode leads to an error:  

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
...
/src/canopy_bioemi_mod.F90:226

Likely this is the case with not having the daytime only limits on calculating biogenic emissions, which is good now especially considering new LDF/LIF impacts and emissions at night, but we will need to have conditions inside of the biogenic emissions module to deal with this.  This is likely the case only in short runs (< 24 hours) using instantaneous leaf PPFD during nighttime hours in certain regions of domain, but need to fix regardless.

drnimbusrain (changes requested):

MaggieMarvin:
Thanks Patrick! I corrected this issue by adding an if statement to control under which conditions to allow gamma PPFD to be calculated: if sunlight is present (dswrf > 0) it will run the calculation, otherwise gamma PPFD is set directly to zero. With these changes, canopy-app compiles and runs with no errors, and it produces the same output as shown above.

drnimbusrain (approved):

Issue #154, opened 2025-03-27
Capturing LAI trends in leaf age calculation


MaggieMarvin:
My understanding of the current leaf age implementation is that the past and current LAI are updated every model time step. In my case I'm running the model hourly with daily LAI and have specified a LAI time step of 86400 s. Does this mean that only once every 24 hours that leaf age is calculated according to an increase or a decrease in LAI? And all other model time steps in between treat the calculation at though past LAI = current LAI?

I understand that LAI is not expected to change much from hour to hour, but is this an appropriate representation of leaf age at a given time of year, when leaf age I think is meant to be more of a seasonal variable?

angehung5:
@MaggieMarvin Maggie, could you please show me which LAI file you are using?

MaggieMarvin:
@angehung5 I believe I pulled them over on Hopper but I don't remember when. All I have now are the input txt files I generated for PEcoRINO and FluCS, but I do see LAI changing on a daily basis. I've started downloading GFS and canopy files from AWS to run monthly simulations and I should use those to regenerate the PEcoRINO/FluCS txt files for consistency, but I think the leaf age calculation will have the same issue either way.

drnimbusrain:
@MaggieMarvin Thanks for bringing this up, as I agree it may not be 100% clear (or consistent compared to MEGAN theory).  

The ```lai_tstep``` in this case is indeed daily (86400 s) since the input LAI data files vary daily (interpolated by @angehung5 from original 8-day VIIRS-NPP product).  Thus, if canopy-app model timestep is < lai_tstep, in this case hourly, yes, LAI should still only change once per day (24 hours).

According to the implementation/formulae though, this seems to be OK if there is no LAI change at hourly time step (i.e., [LAIcurrent = LAIpast](https://github.com/noaa-oar-arl/canopy-app/blob/develop/src/canopy_utils_mod.F90#L703)) until the next hour occurs into the next day.   

@quaz115 based this off of the [MEGANv2 HEMCO implementation](https://github.com/geoschem/HEMCO/blob/main/src/Extensions/hcox_megan_mod.F90#L2302), which I do believe is based on a number of days between LAI values (i.e., days between ```DBTWN```), and in our implementation we technically allow for a sub-daily lai_tstep (but LAI time step cannot be less than model time step), and by default we assume its daily.  

[...]

PR #155, opened 2025-04-24
Correction to LAI tracking for leaf age calculation

MaggieMarvin:
These changes are in response to issue #154 which found that updating past and current LAI with each model time step was leading to errors in the leaf age calculation. Specifically, only one model time step in each LAI time step was capturing the correct change in LAI between past and current conditions, and all other time steps assumed that past LAI = current LAI ([test-1.txt](https://github.com/user-attachments/files/19897316/test.txt): columns are past LAI and current LAI, and rows correspond to each model time step). How past LAI relates to current LAI (<, >, or =) determines how the leaf age calculation is carried out.

Here, past LAI and current LAI are updated once every LAI time step and are saved / carried forward through the intervening model time steps to ensure that the majority time series captures the overall trend in LAI ([test-2.txt](https://github.com/user-attachments/files/19897396/test-2.txt)). For the PEcoRINO case, most model time steps now show correctly that past LAI > current LAI, and the overall effect of the leaf age gamma factor on total biogenic emissions is generally reduced.

Example stats for isoprene before corrections (obs and gfs refer to meteorology used to drive simulations):
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-6c65fe67-7fff-efd1-d192-1eae8395e81b">
leafage_opt* | obs NMB | obs NME | obs r2 | gfs NMB | gfs NME | gfs r2
-- | -- | -- | -- | -- | -- | --
off | −11% | 46% | 0.66 | 49% | 79% | −0.30
on (base) | −16% | 45% | 0.68 | 41% | 73% | −0.09
MEGAN v2/v3 | 138% | 158% | −2.82 | 53% | 74% | 0.24

</b>

And after corrections: 
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-24d39135-7fff-6976-5f8d-3519c7086237">
leafage_opt | obs NMB | obs NME | obs r2 | gfs NMB | gfs NME | gfs r2
-- | -- | -- | -- | -- | -- | --
off | −11% | 46% | 0.66 | 49% | 79% | −0.30
on (base) | −14% | 46% | 0.68 | 49% | 79% | −0.29
MEGAN v2/v3 | 138% | 158% | −2.82 | 53% | 74% | 0.24

</b>

Similar reductions in the overall impact of the leaf age gamma factor are seen for monoterpenes but differences from prior simulations also reflect changes to the calculation of TKA from recent dry dep additions.

### DIFF
--- a/src/canopy_calcs.F90
+++ b/src/canopy_calcs.F90
@@ -32,7 +32,10 @@ SUBROUTINE canopy_calcs(nn)
     !Local variables
     integer i,j,k,loc
     ! LAI variables for Leaf Age factor calculations
-    REAL(rk) :: pastlai             !Past LAI [cm2/cm2]
+    INTEGER  :: int_nlaic           !Int number of LAI timesteps elapsed in the current model timestep
+    INTEGER,  save :: int_nlaip     !Int number of LAI timesteps elapsed in the past model timestep
+    REAL(rk) :: nlaic, nlaip        !Number of LAI timesteps elapsed in past and current model timesteps
+    REAL(rk), save :: pastlai       !Past LAI [cm2/cm2]
     REAL(rk), save :: currentlai    ! Current LAI [cm2/cm2]  (saved from one timestep to the next)
     REAL(rk) :: tsteplai            !Number of days between the past and current LAI
     !Historical Averaging variables for biogenics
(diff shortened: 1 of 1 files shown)


MaggieMarvin:
Also noting here that in prior simulations, I was using a constant LAI with the observed meteorology. But the results in the second stats table reflect the change to run all simulations with the climatological LAI from the GFS/canopy data to ensure that the same trends are captured for the leaf age calculation, no matter which meteorology is used.

drnimbusrain:
Great Maggie!  I can review soon when I return.

drnimbusrain (approved):

PR #158, opened 2025-06-10
Include wetland vtype with grasslands for biogenic parameter assignment

MaggieMarvin:
I noticed this bug when running the gridded set up with DEBUG = 2: the model was crashing with a floating-point exception error traced back to undefined values for ROOTA and ROOTB for vegetation type 11 (wetlands). A closer look at the code showed that EF was also not defined for this vtype and was instead being set to zero in the else statement of the same if else structure.

Since we do expect some biogenic emissions from wetlands, we addressed these issues by lumping wetlands in with grasslands and assigning the same values for biogenic parameters EF, ROOTA, and ROOTB. With this fix under the same model set up the simulation completes successfully with no crash.


drnimbusrain (approved):
Looks good Maggie!

PR #159, opened 2025-06-10
Update README and global python script

angehung5:
- Update LAI and CANFRAC data period in var table. 
- Add user-specified option for canopy data in global python script. Now users can use local canopy files by specifying file location without downloading example daily global gfs files from AWS. Global 0.01 files are used by default.


README.md
drnimbusrain, line 1: @angehung5 This looks good, but technically the 2022 year pertains to the pre-combined gfs.canopy files with GFS 2022 met.  Does your updates to global python canopy script have the ability to use LAI and canfrac from the other 2021-2024 years based on user defined simulation year?  I think a different description would be needed to say that certain years are available for use in LAI and canfrac, not just 2022. This description may also need to be updated below the table too in the AWS discussion sections, as it says things like "_Global 13-km global canopy files (**based on 2020 satellite data**) combined with 2022 GFS meteorology are available on [AWS](https://registry.opendata.aws/noaa-oar-arl-nacc-pds/):_".  Is this still strictly true, or are the updated python scripts adding more functionality?

python/global_data_process.py
drnimbusrain, line 1: @angehung5 This is really great, thanks!  Can you explain more if these updates now allow for user defined specific years of LAI and canfrac to be combined with representative GFS met for canopy-app simulations?  IF so, I think other updates are also needed to python README, e.g., in https://github.com/noaa-oar-arl/canopy-app/blob/develop/python/README.md#generating-global-data  It still says strictly things like "_The daily global canopy data files based on year 2022 and operational NOAA/NWS GFS meteorological files (for the entire record March 23, 2021 - Current; uploaded daily at ~ 18 UTC) are available on [AWS](https://noaa-oar-arl-nacc-pds.s3.amazonaws.com/index.html). If the required NetCDF files are not locally available in the user specified working directory, the python script will automatically download these 2022 canopy data from AWS and merge with the needed GFS meteorological variables for the specific date/times needed to run canopy-app._"  Does this change now with new functionality?

drnimbusrain (changes requested):

README.md
angehung5: The update python script keeps the functionality to use pre-generated daily global files (with 2022 GFS met) while allows users to specify local canopy data files depending on users' needs. So I suggest stating 2022 for LAI and canfrac in the table since the daily global files are still based on year 2022, and mentioning the new functionality in the python README. See my answer below.

python/global_data_process.py
angehung5: Not necessarily changed. The script still can download daily global canopy files based on year 2022, but also can use user-specified local canopy data. Now the script will automatically download daily canopy files only when pre-generated canopy data option is used (can_src=0), so user can have the flexibility to use their own data. The 0.01 global canopy files will serve as a potential data source for users as mentioned in README.

angehung5:
OK, I added one additional option for canopy data. Now users can either use daily global gfs canopy files (can_src=0), get global canopy files from AWS (can_src=1), or specify local canopy files (can_src=2). The two new options have been tested and work properly.

README.md
drnimbusrain: I agree with you Wei-Ting, thanks!

python/global_data_process.py
drnimbusrain: Excellent Wei-Ting, thank you!!

drnimbusrain (approved):
@angehung5 Thank you for this excellent work, which is much more flexible now to use variable varying years for things like LAI, and have also the option for user provided canopy data.  Well done! Approving and merging now!

@MaggieMarvin This is now updated and ready for your testing and paper for generating biogenic simulations and inventory years 2022-2024 hopefully.

PR #165, opened 2025-07-23
Fix github pages deploy action 

bbakernoaa:
The current github action only deployed from the main branch.  Adding options to deploy from both the main and develop branch

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -97,7 +97,7 @@ jobs:
     name: Deploy to GitHub Pages
     runs-on: ubuntu-22.04
     needs: build-docs
-    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -97,7 +97,7 @@ jobs:
     name: Deploy to GitHub Pages
     runs-on: ubuntu-22.04
     needs: build-docs
-    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop') && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```
drnimbusrain:
```suggestion
    if: (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/develop') && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```
Our main branch is called stable, so think this change is needed.

drnimbusrain (changes requested):

drnimbusrain (approved):

